PR #88, opened 2020-09-16
refactor(stack-cntlr): support private git repos using Automation API

### LINKED ISSUE #84: Support for private git repos
The operator seems to currently only support public repos. Is there a way to pass a token in the Stack resource so that private repos can be used?

metral:
### Proposed changes

refactor(stack-cntlr): support private git repos using Automation API

### Related issues (optional)

Closes #84


EvanBoyle (approved):
🐿️  chip it!

PR #106, opened 2020-11-20
Add `backend` configuration

### LINKED ISSUE #83: Add support for storage backend
Pulumi support storing the state in a backend similar to Terraform - AWS S3, GCS etc
The suggestion is to add support for this - allow to set in the secret, instead of access token, the relevant backend:
```
stringData:
  account: "gs://pulumi-state"
```

Looking at the code, this is relatively simple to add - unless there is a reason it is not there from the beginning :) 
This could also support on-prem deployments of pulumi - by setting the `account` key to the relevant server URL.

lukehoban:
Adds a top level `backend` field to the Stack CR that allows pointing all operations for the target Stack at the given backend.  In cases where `backend` is set to a non-service backend, the AccessTokenSecret field is also no longer required.

This is dependent on https://github.com/pulumi/pulumi/pull/5789, and cannot be merged until the operator can take a dependency on that change in the core `pulumi` CLI.

Fixes #83.

### DIFF
--- a/pkg/apis/pulumi/v1alpha1/stack_types.go
+++ b/pkg/apis/pulumi/v1alpha1/stack_types.go
@@ -12,12 +12,15 @@ import (
 type StackSpec struct {
 	// Auth info:
 
-	// AccessTokenSecret is the name of a secret containing the PULUMI_ACCESS_TOKEN for Pulumi access.
-	AccessTokenSecret string `json:"accessTokenSecret"`
+	// (optional) AccessTokenSecret is the name of a secret containing the PULUMI_ACCESS_TOKEN for Pulumi access.
+	AccessTokenSecret string `json:"accessTokenSecret,omitempty"`
 	// (optional) Envs is an optional array of config maps containing environment variables to set.
 	Envs []string `json:"envs,omitempty"`
 	// (optional) SecretEnvs is an optional array of secret names containing environment variables to set.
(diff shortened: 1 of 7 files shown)


stack72 (approved):
This is great - should we document (or point to the backend docs) what backends we support?

pkg/apis/pulumi/v1alpha1/stack_types.go
```diff
@@ -12,13 +12,16 @@ import (
...
 	// Auth info:
 
-	// AccessTokenSecret is the name of a secret containing the PULUMI_ACCESS_TOKEN for Pulumi access.
-	AccessTokenSecret string `json:"accessTokenSecret"`
+	// (optional) AccessTokenSecret is the name of a secret containing the PULUMI_ACCESS_TOKEN for Pulumi access.
+	AccessTokenSecret string `json:"accessTokenSecret,omitempty"`
 	// (optional) Envs is an optional array of config maps containing environment variables to set.
 	Envs []string `json:"envs,omitempty"`
 	// (optional) SecretEnvs is an optional array of secret names containing environment variables to set.
 	SecretEnvs []string `json:"envSecrets,omitempty"`
```
stack72, line 21: nit spacing :D

lblackstone (approved):

viveklak:
@lukehoban You might have to rebase to pick up the GHA fixes.

PR #110, opened 2020-12-01
Get rid of deprecated GHA actions

### LINKED ISSUE #108: Github Actions failing due to use of set-env and get-path commands
### Problem description
Currently GHA workflows are failing due to use of `set-env` and `get-path` commands. See https://github.com/pulumi/pulumi-kubernetes-operator/pull/107/checks?check_run_id=1475352598 for an example.

viveklak:
Fixes #108.

### DIFF
--- a/.github/workflows/push_and_pr.yaml
+++ b/.github/workflows/push_and_pr.yaml
@@ -59,8 +59,7 @@ jobs:
     - name: Set env variables
       run: |
         echo 'PATH=$HOME/.pulumi/bin:$PATH' >> $GITHUB_ENV
-        STACK=ci-cluster-$(head /dev/urandom | LC_CTYPE=C tr -dc '[:lower:]' | head -c5)
-        echo "$STACK" >> $GITHUB_ENV
+        echo "STACK=ci-cluster-$(head /dev/urandom | LC_CTYPE=C tr -dc '[:lower:]' | head -c5)" >> $GITHUB_ENV
     - name: Tests
       run: |
         # Create GKE test cluster to install CRDs and use with the test operator.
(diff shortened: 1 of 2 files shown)


.github/workflows/push_and_pr.yaml
```diff
@@ -56,13 +56,13 @@ jobs:
       run: |
         # Install Pulumi
         curl -fsSL https://get.pulumi.com | bash -s
-        export PATH=$HOME/.pulumi/bin:$PATH
-        echo "::set-env name=PATH::$PATH"
+    - name: Set env variables
+      run: |
+        echo 'PATH=$HOME/.pulumi/bin:$PATH' >> $GITHUB_ENV
```
stack72, line 61: this should be GITHUB_PATH - it's not an env var

.github/workflows/tag_release.yaml
```diff
@@ -38,13 +38,13 @@ jobs:
       run: |
         # Install Pulumi
         curl -fsSL https://get.pulumi.com | bash -s
-        export PATH=$HOME/.pulumi/bin:$PATH
-        echo "::set-env name=PATH::$PATH"
+    - name: Set env variables
+      run: |
+        echo 'PATH=$HOME/.pulumi/bin:$PATH' >> $GITHUB_ENV
```
stack72, line 43: same >> GITHUB_PATH

.github/workflows/tag_release.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v2
     - name: Install Go
-      uses: actions/setup-go@v2.1.1
+      uses: actions/setup-go@v2.1.3
```
stack72, line 16: if we remove the .1.3 and make it just actions/setup-go@v2 then we get the minors for free

stack72 (changes requested):
Changes inline

stack72 (approved):
Thank you!!!

Issue #118, opened 2021-02-21
Provide CRD and resources as library


eljoth:
The workflow I am expecting from the Pulumi operator is as follows:
1. Create a repository with a working Pulumi program, containing the Pulumi operator.
2. Deploy the project to a cluster manually one time.
3. Let the Operator handle changes in desired state automatically (see #103)

Currently, the developer has to either deploy the crd on his own by applying the yaml or translate the existing yaml into his pulumi program. In both cases it needs unnecessary effort to keep the Operator updated by manually changing those files and keep them updated.

## Affected feature

Provide the CRD and all resources required as library (e.g. Go Module) to make as well deploying as upgrading easier. Developers just update the dependency and take care of changes breaking their code and all they have to do  is push the changes to their repo. Pulumi Operator will notice that there are changes and redeploy itself in an updated version. This feature enables a Helm-like experience with the addition of all advantages Pulumi has to offer.

eljoth:
Maybe we could make use of [crd2pulumi](https://github.com/pulumi/crd2pulumi) to automate this procedure?

UnstoppableMango:
Not sure if it's been mentioned elsewhere, but the CRDs don't seem to map cleanly when using `crd2pulumi` because of the name clashes with the `Pulumi`/`pulumi` namespace

PR #119, opened 2021-02-25
Provide pulumi operator as embeddable library

eljoth:
### Proposed changes

This PR wants to add the functionality to provide the pulumi operator as a library to be used in pulumi programs. It offers a custom component resource containing all dependencies for the operator including the deployment of it and provides  pulumi bindings for the stack crd.

### Related issues (optional)

#118


lukehoban:
Thanks @eljoth the the PR!

This is a great idea to add these SDKs.  

Am I correct in assuming these were created with `crd2pulumi`?

A few high-level notes:
1. We should include Makefile build steps that recreate these SDKs 
2. We should include publishing for these SDKs to the various package managers (@stack72 might be able to help with this)
3. Depending on how we're going to publish this - that might inform what namespace this actually belongs in.  Currently it appears to be in a top-level `pulumi` namespace - which likely won't work?
4. I wonder if it might make sense to add this to https://github.com/pulumi/pulumi-kubernetes-crds instead of homing/publishing it here?
5. Would be great to add a note on these to the README.
6. It would be great to add an example of using this e2e - deploying the operator and then managing a `Stack` using these SDKs, all within Pulumi!

eljoth:
@lukehoban yes, all of them were created with crd2pulumi. In addition, I implemented the role, rolebinding and deployment by hand. 

I will continue in the next few days to update the makefile and other files and publish the SDKs to the package managers. Regarding the publishing I would of course appreciate any help!
As of what I've seen in [pulumi-kubernetes-crds](https://github.com/pulumi/pulumi-kubernetes-crds/tree/master/operators/splunk/go/enterprise/v1alpha2), the sources "only" seem to provide the CRDs but not a way to actually deploy the operator itself. I don't know whether it was a better choice to keep it all together here or to split things up.

lukehoban:
> As of what I've seen in pulumi-kubernetes-crds, the sources "only" seem to provide the CRDs but not a way to actually deploy the operator itself. I don't know whether it was a better choice to keep it all together here or to split things up.

Yes - I believe that is right - and I agree that it is nice to keep it together.  Ideally the operators in pulumi-kubernetes-crds would also include the Pulumi resources needed to deploy the operators themselves.  But for now - makes sense to me to keep there here.  

+ also @lblackstone and @justinvp.

PR #138, opened 2021-04-29
Distinguish debug and info logging

### LINKED ISSUE #140: Operator logs sensitive information on info level
The operator allows reading various configuration items via `ResourceRef` which is great. However, upon retrieving the configs from the Stack CR it logs all values on `info` causing secrets leakage.

## Expected behavior

If these configs need to be logged at all, either mask these values before doing so or increase logging verbosity to `debug`.

## Current behavior

Secrets leak on `info`. Also the operator default manifests set the level to `debug`, see https://github.com/pulumi/pulumi-kubernetes-operator/blob/master/deploy/yaml/operator.yaml#L23

## Steps to reproduce

1. Deploy operator
2. Deploy stack

```
apiVersion: pulumi.com/v1alpha1
kind: Stack
metadata:
  name: stack
  namespace: pulumi-operator
spec:
  commit: xzy
  envRefs:
    PULUMI_ACCESS_TOKEN:
      filesystem:
        path: /path/to/pulumi_api_token
      type: FS
...
```

3. Observe log emitting all stack properties on `info`.

viveklak:
Fixes #140

### DIFF
--- a/pkg/controller/stack/session_test.go
+++ b/pkg/controller/stack/session_test.go
@@ -3,6 +3,7 @@ package stack
 import (
 	"errors"
 	"fmt"
+	"github.com/pulumi/pulumi-kubernetes-operator/pkg/logging"
 	"io/ioutil"
 	"os"
 	"testing"
@@ -51,7 +52,7 @@ func TestSuite(t *testing.T) {
 
 func (suite *GitAuthTestSuite) TestSetupGitAuthWithSecrets() {
 	t := suite.T()
-	logger := log.WithValues("Request.Test", "TestSetupGitAuthWithSecrets")
+	logger := logging.NewLogger(t.Name(), "Request.Test", "TestSetupGitAuthWithSecrets")
 
 	sshPrivateKey := &corev1.Secret{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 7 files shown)


pkg/controller/stack/session_test.go
```diff
@@ -3,6 +3,7 @@ package stack
 import (
 	"errors"
 	"fmt"
+	"github.com/pulumi/pulumi-kubernetes-operator/pkg/logging"
```
elsesiy, line 6: Given the existing ordering, this import should be on L12 if I'm not mistaken

pkg/controller/stack/stack_controller.go
```diff
@@ -7,14 +7,15 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/pulumi/pulumi-kubernetes-operator/pkg/logging"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
```
elsesiy, line 11: Same comment as above, these are not stdlib imports

pkg/logging/logger.go
```diff
@@ -0,0 +1,80 @@
+package logging
```
elsesiy, line 1: I just noticed that none of the files have the preamble such as https://github.com/pulumi/pulumi/blob/master/sdk/go/auto/stack_test.go#L1

pkg/controller/stack/session_test.go
viveklak: We typically adhere to `go fmt` which is admittedly a little weird about it. Still prefer it to `go imports` which does some arbitrary newline sprinkling.

pkg/logging/logger.go
viveklak: Yeah we recently added a check for this in pulumictl https://github.com/pulumi/pulumictl/tree/master/cmd/pulumictl/copyright. This repo hasn't been hooked up to it yet. Will do that separately.

lblackstone (approved):

PR #156, opened 2021-05-25
Bump pulumi/pulumi and add useragent string

### LINKED ISSUE #155: Set Automation API `UserAgent` option
We should be setting user agent https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3@v3.3.0/go/auto/optup#UserAgent when doing an up/preview/destroy/refresh. This metadata gets persisted in backends so users can differentiate between updates driven by CLI, or other automated tools such as GHA, the operator, etc.

viveklak:
Fixes #155

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -2,7 +2,7 @@ project_name: pulumi-kubernetes-operator
 
 builds:
   - ldflags:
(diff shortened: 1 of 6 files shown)


EvanBoyle (approved):
🚢

PR #186, opened 2021-08-25
Add useLocalStackOnly option to prevent stack creation

### LINKED ISSUE #185: Operator assumes stack creation if it doesn't exist
The current implementation always assumes that a stack that doesn't exist in the tracking git repo to be automatically created, see [here](https://github.com/pulumi/pulumi-kubernetes-operator/blob/master/pkg/controller/stack/stack_controller.go#L620).

This could be problematic as users might want to decouple the stack creation from the stack CR creation (often live in different repos). Stacks often require certain configs to be present and assume the default values as part of the created stack by the stack CR might not be desirable.

elsesiy:
Also bump pulumi/pulumi to 3.10.3

Fix #185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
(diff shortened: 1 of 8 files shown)


elsesiy:
cc @clstokes @viveklak

viveklak (approved):

PR #195, opened 2021-09-14
Improve workdir cleanup logic

### LINKED ISSUE #104: Ever growing collection of stack and ssh directories in /tmp

viveklak:
Fixes #104.

Manually verified that the work directory is deleted on failure while retaining existing behavior on success.

### DIFF
--- a/pkg/controller/stack/stack_controller.go
+++ b/pkg/controller/stack/stack_controller.go
@@ -174,13 +174,13 @@ func (r *ReconcileStack) Reconcile(ctx context.Context, request reconcile.Reques
 		return reconcile.Result{}, err
 	}
 
+	// Delete the working directory after the reconciliation is completed (regardless of success or failure).
+	defer sess.CleanupPulumiWorkdir()
+
 	currentCommit, err := revisionAtWorkingDir(sess.workdir)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-
-	// Delete the working directory after the reconciliation is completed (regardless of success or failure).
-	defer sess.CleanupPulumiWorkdir()
 
 	// Step 2. If there are extra environment variables, read them in now and use them for subsequent commands.
 	if err = sess.SetEnvs(stack.Envs, request.Namespace); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/stack/stack_controller.go
```diff
@@ -656,7 +672,7 @@ func (sess *reconcileStackSession) SetupPulumiWorkdir(gitAuth *auto.GitAuth) err
 		return errors.Wrap(err, "installing project dependencies")
 	}
 
-	return nil
+	return err
```
mikhailshilkov, line 675: I can't quite grasp this change...

pkg/controller/stack/stack_controller.go
```diff
@@ -599,7 +599,22 @@ func (sess *reconcileStackSession) SetupPulumiWorkdir(gitAuth *auto.GitAuth) err
...
+
+	// Cleanup the workdir on failure setting up the workspace.
+	defer func() {
+		if err != nil {
+			_ = os.RemoveAll(dir)
+		}
+	}()
+
+	var w auto.Workspace
+	w, err = auto.NewLocalWorkspace(context.Background(), auto.WorkDir(dir), auto.Repo(repo), secretsProvider)
```
mikhailshilkov, line 617: What's the default behavior of our automation API before this change? Does it create a temporary folder internally but does not clean it up correctly? What is our advice to 3rd party users of automation API?

pkg/controller/stack/stack_controller.go
viveklak: > I can't quite grasp this change...

Since the defered function will check the value of `err`, the change makes sure returned value is through err. Could have done the same with a named return value. Happy to use that if that makes things clearer.

pkg/controller/stack/stack_controller.go
viveklak: > What's the default behavior of our automation API before this change? Does it create a temporary folder internally but does not clean it up correctly? What is our advice to 3rd party users of automation API?

Yes it creates a temporary directory internally. We tried to handle cleanup through the `sess.CleanupPulumiWorkdir()` call elsewhere but we do many operations as part of the stack setup which may fail independently and we leak the work dir on failure.
This seems to be more a function of the way the operator is currently written than anything specific to the automation API.

lblackstone (approved):

pkg/controller/stack/stack_controller.go
viveklak: > I can't quite grasp this change...

Its actually unnecessary. Removed.

mikhailshilkov (approved):

PR #249, opened 2022-02-01
fix: use runtime namespace when available

rawkode:
When running the operator in a namespace, stack resources that do not contain an explicit namespace are attempted to be created in the default namespace; which is rather strange behaviour.

This PR will infer the default namespace from the service account under which the pod is running.

### Proposed changes

This change has the operator set the default namespace to that of the service account under which it's running.

### Related issues (optional)

N/A

### DIFF
--- a/pkg/controller/stack/stack_controller.go
+++ b/pkg/controller/stack/stack_controller.go
@@ -1205,6 +1205,8 @@ func contains(list []string, s string) bool {
 func setupInClusterKubeconfig() error {
 	const certFp = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 	const tokenFp = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+	const namespaceFp = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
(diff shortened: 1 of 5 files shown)


pkg/controller/stack/stack_controller.go
```diff
@@ -1229,14 +1235,15 @@ contexts:
 - context:
     cluster: local
     user: local
+    namespace: %s
```
viveklak, line 1238: Should we add an environment variable to potentially disable this? Agree with this being the default though.

rawkode: How about `PULUMI_INFER_NAMESPACE` ?

pkg/controller/stack/utils.go
```diff
@@ -0,0 +1,23 @@
+package stack
```
viveklak, line 1: Add copyright

viveklak (approved):
Thanks so much! Could you also add the environment variable to https://github.com/pulumi/pulumi-kubernetes-operator/blob/master/deploy/operator_template.yaml#L44 to enable this behavior by default?

viveklak (approved):

PR #282, opened 2022-06-01
Guard deref of pointer field .Status.LastUpdate

squaremo:
In a new resource, the status will be the zero value, meaning pointer-valued fields will be nil. The controller dereferences the field `.Status.LastUpdate` without first checking whether it's nil, leading to a nil pointer dereference.

The quick solution in this case is to skip the offending block if the field is nil. This avoids the nil pointer dereference; and it's correct because the code within the block is irrelevant if there's no last update.

### DIFF
--- a/pkg/controller/stack/stack_controller.go
+++ b/pkg/controller/stack/stack_controller.go
@@ -272,7 +272,7 @@ func (r *ReconcileStack) Reconcile(ctx context.Context, request reconcile.Reques
 		}
 	}
 
-	if trackBranch {
+	if trackBranch && instance.Status.LastUpdate != nil {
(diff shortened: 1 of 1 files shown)


squaremo:
If you are wondering "How did this ever work?": in the normal course of things, the method `addDefaultPermalink` would be called _before_ the unguarded dereference, and `addDefaultPermalink` initialises the `.Status.LastUpdate` field if it's nil. I encountered the NPE, so clearly this doesn't always happen, and in any case, it's something of a coincidence.

stack72:
Hi @squaremo 

Thanks for finding this :)

P.

viveklak (approved):

PR #293, opened 2022-06-30
Move deletion handling up front

squaremo:
At present, the git repo fields in the stack spec are validated only if
the resource is not marked to be deleted, but further work using those
fields is done unconditionally. This is incorrect, since invalid field
values might be used.

If the resource is due to be deleted, there is nothing else to do, so
the controller should just clean up and exit early (since the
finalizer-adding code goes with that, it moves too). Then the
validation, and further work if validation succeeds, can proceed
unconditionally.

### DIFF
--- a/pkg/controller/stack/stack_controller.go
+++ b/pkg/controller/stack/stack_controller.go
@@ -147,6 +147,13 @@ type ReconcileStack struct {
 	recorder record.EventRecorder
 }
 
+func validateGitRepo(repo shared.StackSpec) error {
+	if repo.Commit == "" && repo.Branch == "" {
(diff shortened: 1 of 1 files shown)


squaremo:
> the controller should just clean up and exit early

It occurs to me after a test failure that in general it will need a git checkout to destroy a stack as part of cleanup. So there's two alternatives:
 1. make the cleanup code clone the git if it needs to; or,
 2. wait until after cloning the repo to check whether cleaning up is the thing we're doing (i.e., the status quo, or near enough).

It think 1) is the better of those, because I think the cleanup code should get the code at the last applied commit, rather than whatever would have been fetched. But that's a subtle consideration; the other reason for 1) is that branching on deletion early is easier to understand when reading the code, and there needn't be duplication of code to achieve it.

PR #315, opened 2022-08-24
Prep v1.8.0 release

viveklak:


CHANGELOG.md
```diff
@@ -4,11 +4,13 @@ CHANGELOG
 ## HEAD (Unreleased)
 (None)
 
+## 1.8.0 (2022-08-24)
 - Use go 1.18 for builds
 - CI update to go 1.18
 - Bump to v3.36.0 of Pulumi, to support short branch names in .spec.branch after
```
squaremo, line 10: I didn't go far enough with #10116 -- branches other than the default are still not supported. In https://github.com/pulumi/pulumi/pull/10285 I extended the support to all branches, so it's worth waiting until a release of pulumi/pulumi that includes it. (Unless other things here are urgent)

squaremo: If urgent -- perhaps leave that entry from the changelog, so we don't get people's hopes up about branch names.

viveklak: > If urgent -- perhaps leave that entry from the changelog, so we don't get people's hopes up about branch names.

I think its ok to wait till a pu/pu release since that seems like a much requested improvement.

stack72 (approved):
Let's change the date and bump to pu/pu before we release :)

CHANGELOG.md
stack72: This is now at pu/pul 3.39.0

PR #318, opened 2022-08-30
Upgrade to Pulumi v3.38.0

stack72:


stack72:
This is actually managed by #315

PR #319, opened 2022-08-31
Isolate envtest tests to local processes, rather than requiring a cluster

squaremo:
### Proposed changes

This changes the envtest (Kubernetes environment for unit testing) configuration so that it runs an API server against etcd in a local process, rather than expecting an existing cluster. This has a couple of benefits:

 - coincidences and flakes are fewer, because the whole environment is created fresh each time
 - no chance of conflicting with a controller running in the cluster
 - there's no need to spin up a cluster in a cloud platform (or use e.g., `kind`) for integration tests

The downsides are:

 - there's a startup time associated with running etcd and kube-api-server each test run, which you may not have if using e.g., Docker Desktop
 - if you rely on other controllers to be running for a test to succeed -- e.g., the replicaset controller creating pods, so that a Deployment reaches a ready state -- that may have to be faked with a Kubernetes client. (On the other hand, usually you would want this to be explicit in your test, anyway)

This PR also makes test fail much faster, when they fail, and fixes up a couple of things that were exposed by tests timing out:
 - a goroutine leak
 - a race hazard between foreground deletion and the controller's finalizer handling


stack72 (approved):
Thank you for this :) the original setup was very much cut and paste from other sources so this makes it more targetted

squaremo:
(Just rebased after #322, since there were adjacent changes)

roothorp (approved):

Issue #299, opened 2022-07-15
Mechanism to refresh dynamic credentials for providers


viveklak:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
In many production scenarios, cloud provider credentials may be retrieved from external entities like Vault. Refresh doesn't instantiate the pulumi program and providers are instantiated from state which might mean using stale credentials if the dynamism is encoded in the Pulumi program. CLI users have mechanisms such as https://github.com/pulumi/pulumi/issues/4981#issuecomment-655881330 to workaround this but this is not supported in autonomous systems like the Operator. This prevents operator users from using mechanisms like auto-refresh added in https://github.com/pulumi/pulumi/pull/8071 for their stacks and might require manual intervention for refreshing credentials.

Approaches discussed so far are to add support one of the following in the stack specification:
1. Support a pre-flight script which can be used to fetch credentials from external entities and populate things like files, environment variables etc. supported by the underlying cloud provider. 
2. Support a mechanism to allow stack authors to selectively target specific URNs for update before running refresh. This is essentially the equivalent of running `pulumi up -t <provider>` as in the CLI workaround mentioned above. With this the credential refresh code could live within the pulumi program and leverage the pulumi sdk instead of in a hand-coded script.

squaremo:
Some additional qualifications:

 - this is an issue with first-class providers specifically, because the problem arises when you explicitly construct a provider in order to do further work; since a constructed provider is a resource, its inputs are captured in the stack.
 - any solution must account for multiple "hops"; that is, must work when a provider is used to construct another provider. 

**Pre-flight script**

If this is an arbitrary shell script, you have to make sure the container image has all the executables that might be invoked in the script. That's API clients for all the cloud platforms, clients for other tooling like Vault. The list is endless, so in practice is means some combination of dumping things in the released container image to cover the 80% needs, and people having to construct their own bespoke images.

**Targeted `pulumi up`**

I would expect this to put some restrictions on the way the program is written (and I would love to understand in detail what they are! :-)

If it is onerous to write a program which serves for `pulumi up -r` as well as `pulumi -t` for providers, perhaps it would be better to have two separate programs, one for doing provider construction, and one for using the providers; and a mechanism for making the latter program depend on the former.

squaremo:
[...]

PR #320, opened 2022-09-07
Ability to update provider state before running refresh

squaremo:
Addresses #299.

 - [x] Add a test case that verifies the "stale state" problem described in #299.
   - [x] (Remove focus from that test case, so others will run!)
 - [x] Write a design for the solution and implement it
   - ~[ ] get the design reviewed and iterate on it~
 - [x] Adapt test case to demonstrate that can be used to solve the stale provider state problem
 - [x] Write a worked example of how to use it

EDIT: remove design review task, which is just not going to happen; add worked example task, since that was specifically requested.


docs/design/prerequisites-and-targets.md
```diff
@@ -0,0 +1,117 @@
...
+    unless:
+      updatedWithinDuration: 1m
+```
+
+If the referenced stack has not been run in the last `updatedWithinDuration` duration, it will be
+queued, and the current stack requeued.
+
+**Q. How does the controller queue the prerequisites?**
+
+One mechanism is to simply add an annotation to the object, which will trigger an update event and
```
viveklak, line 54: If we are using the drift-detection mode (`ContinueResyncOnCommitMatch`) we wouldn't need to tickle the stack right? I imagine we could align this with `updatedWithinDuration`.

docs/design/prerequisites-and-targets.md
```diff
@@ -0,0 +1,117 @@
...
+that is how you use it to e.g., update rotated credentials, if that's done all in one program.
+
+### Targets
+
+The field `.spec.targets` has a list of URNs to supply to the Pulumi automation API when running the
+stack.
+
+## How this design solves the stale provider state problem
+
+To solve the stale provider state problem, you would make two stacks:
```
viveklak, line 79: One challenge - whether here or through one of the alternatives discussed later is to know ahead of time the URNs to target. Making this a prerequisite for the regular stack execution is going to introduce a circular dependency no? As in - the provider URN to target doesn't exist until the full stack is executed?

squaremo: Yes -- this is a problem for this workaround, wherever you do it. I can think of a few improvements:

 - instruct people to make the "main" stack first, extract the URN of the provider, then set up the refresh stack. Since the problem doesn't exist unless you have run the stack at least once, there's no circularity, just a step sideways.
 - ask for something other than URNs, and try to figure out what the URNs will be. This _might_ end up being circular, or at least fiddly.

docs/design/prerequisites-and-targets.md
squaremo: Right, you could just let the prerequisite get run in the normal course of things, and come back later. There's a chance the schedules won't match well, and you get syncopated updates -- that is, you sometimes or always miss your `updatedWithinDuration` window because the resync period is too long. In that case, explicitly queueing prerequisites may be necessary for correctness.

docs/design/prerequisites-and-targets.md
viveklak: I wonder if it would make sense to model the dependency aspect as a separate CRD - e.g. `stacklink` - something that links dependent stacks together. This essentially looks like a dependency link between stacks and feels a bit awkward to shove this into the stack spec? In my view the prerequisite - and target seem like constraints placed on a stack which can be more dynamic in nature. We may be able to provide a simple expression to resolve URNs etc. over time to make this easier. 

In some ways there are echoes of https://github.com/pulumi/pulumi/issues/2309 here which is a much requested feature and could enable a number of the kind of workloads mentioned there over time.

cc @lukehoban for thoughts as well.

squaremo: > This essentially looks like a dependency link between stacks and feels a bit awkward to shove this into the stack spec?

I don't think it's awkward -- there are already a bunch of fields that are to do with when and how a Stack object is processed (rather than what it executes), like `RetryOnUpdateConflict` and `ContinueResyncOnCommitMatch`. This is just another one of those.

> In some ways there are echoes of https://github.com/pulumi/pulumi/issues/2309

For sure; if `StackReference` dependencies could be chased down and updated, it would give you a natural way to write programs that didn't need the workaround here. I'm not sure it would cover every use case*, but to me it's the right primitive to start with.

*Why not? Hand waving a bit,
 - using a stack reference means you have to hand values from one program to another via outputs, and you may not want intermediate values to be outputs;
 - with a stack reference, you don't get to say how the referenced stack is executed (e.g., you can't provide targets)

squaremo: The main benefit of having another type to represent dependence is that it makes it possible to separate the ownership of the stack definition and its dependence on other stacks. This could be important when crossing responsibility boundaries, e.g., a platform team can make app stacks depend on platform infrastructure, without the app teams needing to know about it.

squaremo:
The image docker.io/squaremo/pulumi-kubernetes-operator:e5b090d is built from the HEAD of this PR branch.

PR #329, opened 2022-09-28
Deprecate cross-namespace refs

squaremo:
Cross-namespace refs won't work with the provided configuration, which gives RBAC permissions for one namespace only. Granting permissions to other namespaces means they are also accessible to the programs run by stacks, which is a security hole.

Therefore: guard the use of `api/pulumi/shared/SecretRef.Namespace`, and watching >1 (or all) namespaces with `WATCH_NAMESPACE`, with an environment variable `INSECURE_NO_NAMESPACE_ISOLATION`. This must be set to `1` or `true` to enable cross-namespace operation.

This is a breaking change for anyone running the operator with either an empty value, or a comma-separated list (e.g., `ns1,ns2` for `WATCH_NAMESPACE`; and, for anyone using cross-namespace secret references (i.e., with a value in `.namespace` under one of the fields listed below). The documented means of installing the operator sets `WATCH_NAMESPACE` to be the namespace in which the operator is deployed.

References in Stack which can be cross-namespace (`*` indicates there's a list or map):
 - `.spec.envRefs[*].secret`
 - `.spec.secretRefs[*].secret`
 - `.spec.gitAuth.accessToken.secret`
 - `.spec.gitAuth.sshAuth.sshPrivateKey.secret`
 - `.spec.gitAuth.sshAuth.password.secret`
 - `.spec.gitAuth.basicAuth.userName.secret`
 - `.spec.gitAuth.basicAuth.password.secret`

To fix the breakage:

 - if you are using a cross-namespace reference, consider using a secret in the same namespace as the Stack object. [Kyverno](https://kyverno.io/policies/other/sync_secrets/sync_secrets/) and other such automation can help with syncing secrets across namespaces.
 - if you use `WATCH_NAMESPACE` with an empty value or several values, modify your configuration so that you're running the operator in each namespace where you have Stack objects. ([This PR](https://github.com/pulumi/pulumi-kubernetes-operator/pull/328) extends the provided Pulumi programs so they will do this for you.)
 - If it is not possible to use one of the fixes above, you can enable watching multiple namespaces and cross-namespace references by adding the env entry `INSECURE_NO_NAMESPACE_ISOLATION` to the operator deployment spec, in the same place as `WATCH_NAMESPACES` is defined. Be aware that this is really only suitable for "single tenant" clusters -- when you are happy for anyone with access to have access to everything.


pkg/controller/stack/stack_controller.go
```diff
@@ -199,18 +255,33 @@ func (r *ReconcileStack) Reconcile(ctx context.Context, request reconcile.Reques
...
+	// indicate a retry is needed (RetryError), processing has stalled (StallError), or a crash
+	// (anything else).
 	saveStatus := func() {
-		if reterr == nil {
-			instance.Status.ObservedGeneration = instance.GetGeneration()
-		} else {
-			// controller-runtime will requeue the object, so reflect that in the conditions by
-			// saying it is still in progress.
-			instance.Status.MarkReconcilingCondition(pulumiv1.ReconcilingRetryReason, reterr.Error())
+		if reterr != nil {
```
lukehoban, line 262: This error handling appears to introduce new semantics around how/when retires and stalls are triggered. Is this strictly related to the namespace isolation changes?  Are we confident these changes are well covered in tests?

lukehoban (approved):
Namespace isolation changes LGTM. One question on the refactoring of reconciliation approach to retries/stalls.

pkg/controller/stack/stack_controller.go
squaremo: Re test coverage: #316 included tests that _some_ arbitrary stalling problem resulted in status conditions indicating a stall, and _some_ arbitrary retryable problem resulted in a status conditions indicating "in progress". This PR has tests that with namespace isolation, using a cross-namespace ref results in a stall, and with namespace isolation waived, it doesn't (all else going well).

For _specific_ stalling or retryable problems -- I'll leave those to the other, hypothetical PR. They could have been individually tested here, I was relying on natural induction :-)

squaremo: The more recent attempt uses the same error type protocol, but only for cross-namespace refs causing a stall, and only at the site of calling the helper; so, everything has the same retry/stall semantics.

yoyoraso:
Hi, I have a question related to the namespace isolation is it possible to deploy one pulumi operator in one namespace to watch and create stacks across the cluster in all namespaces ?

PR #331, opened 2022-10-04
Parallelise and randomize tests

squaremo:
Many of the tests take O(1m) to run, and the more specs added the longer the runtime. Parallelising the tests will help defer the day we need to increase the overall timeout.

Randomising helps keep the tests independent, by surfacing things that work by coincidence.

Enabling either of these requires the use of the Ginkgo command line -- it's not possible to enable them programmatically. So: reinstate the use of `ginkgo` for running tests.

(Empirically, I found that using `ginkgo -p` would result in timeouts waiting for the control plane to start. `-nodes=4` seems like a reasonable compromise.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,8 @@ push-image:
 	docker push $(IMAGE_NAME):$(VERSION)
 
 test: codegen download-test-deps
-	KUBEBUILDER_ASSETS="$(shell setup-envtest --use-env use -p path)" go test -v ./test/...
+	KUBEBUILDER_ASSETS="$(shell setup-envtest --use-env use -p path)" \
+		ginkgo -nodes=4 --randomizeAllSpecs ./test/...
 
 deploy:
 	kubectl apply -f deploy/yaml/service_account.yaml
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -45,7 +45,8 @@ push-image:
 	docker push $(IMAGE_NAME):$(VERSION)
 
 test: codegen download-test-deps
-	KUBEBUILDER_ASSETS="$(shell setup-envtest --use-env use -p path)" go test -v ./test/...
+	KUBEBUILDER_ASSETS="$(shell setup-envtest --use-env use -p path)" \
+		ginkgo -nodes=4 --randomizeAllSpecs ./test/...
```
lukehoban, line 49: Is this now also a dev machine dependency?  Should it be added to README/CONTRIBUTING?

lukehoban (approved):

Makefile
squaremo: `ginkgo` is mentioned as a prerequisite in `docs/build.md` already (I didn't remove it when I changed the Makefile [the last time](https://github.com/pulumi/pulumi-kubernetes-operator/pull/319)).

PR #332, opened 2022-10-05
Tidy git repo source fields into an inlined struct

squaremo:
To prepare the way for referring to other sources of Plumui programs in the Stack spec, this commit moves the fields specifying how to fetch from a git repo directly into their own struct. It's not possible to change the _schema_ in a backward-compatible way, so the struct is serialised inline and won't look any different in the API.

There is a semantic change: the field `.spec.projectRepo` becomes optional (the others involved were already optional). The controller will now mark a Stack object as stalled if it does not specify a source at all, or if it has a "git source" but an empty `projectRepo`.

~This should remain a draft (or at least unmerged) until there's a release not including it.~ We can base the branches for #324 and #326 (or its implementation) on this branch though, to soften the rebase later.


roothorp (approved):
Seems straightforward to me! Is you intent to build the flux integration into the GitSource struct?

PR #340, opened 2022-10-14
Make mutual exclusion of sources obvious

squaremo:
This changes the which-source-is-it switch around a little, so that instead of each case checking something is non-nil and all the others are nil, the invalid cases are ruled out first. This makes it a little easier to extend, since you only have to update the constraint check, then add your case in.

Logical check for invalidity:

```go
(gitPresent == fluxPresent && fluxPresent == programPresent) ||  // i.e., all true or all false
		(gitPresent && fluxPresent) ||                   // ) or, any two are true
		(fluxPresent && programPresent) ||               // ) ...
		(gitPresent && programPresent)                   // ) ...
```

This is a bit unwieldy with three cases, and would be unmanageable with four or more; so I wrote this iterative check:

```go
exactlyOneOf := func(these ...bool) bool {
	var found bool
	for _, b := range these {
		if found && b {
			return false
		}
		found = found || b
	}
	return found
}
```

### DIFF
--- a/pkg/controller/stack/stack_controller.go
+++ b/pkg/controller/stack/stack_controller.go
@@ -270,19 +270,27 @@ func (r *ReconcileStack) Reconcile(ctx context.Context, request reconcile.Reques
 
 	// Step 1. Set up the workdir, select the right stack and populate config if supplied.
 
+	exactlyOneOf := func(these ...bool) bool {
+		var found bool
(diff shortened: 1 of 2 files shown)


roothorp (approved):

squaremo:
I've rebased to remove two commits that were already part of #342.

PR #360, opened 2022-10-31
Use test helpers wherever they can be used

squaremo:


roothorp (approved):
Love it! Super clean.

PR #362, opened 2022-11-01
Examples for Inline Programs (Pulumi YAML)

roothorp:
Add some examples for creating inline programs with Pulumi YAML.


squaremo:
What does this need in order to come out of draft?

PR #384, opened 2022-12-19
readme: Fix the pulumi new instructions

stevesloka:
### Proposed changes

Fixes the pulumi new instructions to fix the template url to be the correct name which caused the example instructions to fail.

Signed-off-by: Steve Sloka <steve@pulumi.com>


squaremo (approved):
My bad -- thanks for finding and fixing this Steve 👍

PR #388, opened 2022-12-20
linter: Adds golangci-lint

stevesloka:
### Proposed changes

Adds golangci-lint to project via Make task and Github Action.

Note: Some tests are still failing and the linter catches, but are resolved in #387

Signed-off-by: Steve Sloka <steve@pulumi.com>

### Related issues (optional)

Updates #387

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -7,6 +7,27 @@ env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
+  lint:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.X
+      - uses: actions/checkout@v3
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
+          version: v1.50
+          # Optional: working directory, useful for monorepos
+          # working-directory: somedir
+
+          # Optional: golangci-lint command line arguments.
+          # args: --issues-exit-code=0
+
+          # Optional: show only new issues if it's a pull request. The default value is `false`.
+          # only-new-issues: true
   comment-on-pr:
     if: github.event.pull_request.head.repo.full_name != github.repository
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


.github/workflows/pull-request.yml
```diff
@@ -19,3 +40,5 @@ jobs:
             PR is now waiting for a maintainer to run the acceptance tests. This PR will only perform build and linting.
             **Note for the maintainer:** To run the acceptance tests, please comment */run-acceptance-tests* on the PR
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+Expect(err):
```
squaremo, line 43: This looks like copypasta (or I don't understand github actions as well as I thought)

test/flux_source_test.go
```diff
@@ -85,7 +85,8 @@ var _ = Describe("Flux source integration", func() {
 		kubectl.Env = append(kubectl.Environ(), fmt.Sprintf("KUBECONFIG=%s", kubeconfig))
 		kubectl.Stdout = os.Stdout
 		kubectl.Stderr = os.Stderr
-		kubectl.Run()
+		err := kubectl.Run()
+		Expect(err).NotTo(HaveOccurred())
```
squaremo, line 89: `ExpectWithOffset` would have the marginal advantage of reporting the caller (assumed to be the test itself) as the location of the failure, rather than this helper.

test/flux_source_test.go
```diff
@@ -99,7 +100,8 @@ var _ = Describe("Flux source integration", func() {
 		// BeforeSuite so it's near where it's used.
 		createCRD.Do(func() {
 			// tell the client about CRDs
-			apiextensionsv1.AddToScheme(k8sClient.Scheme())
+			err := apiextensionsv1.AddToScheme(k8sClient.Scheme())
```
squaremo, line 103: `Expect(<reasonably short invocation>).To(Succeed())` is kind of the "house style", though admittedly not used everywhere.

Makefile
```diff
@@ -50,6 +50,9 @@ test: codegen download-test-deps
 	KUBEBUILDER_ASSETS="$(shell setup-envtest --use-env use -p path)" \
 		ginkgo -nodes=${TEST_NODES} --randomize-all ./test/...
 
+lint:
+	golangci-lint run --timeout 10m --build-tags integration
```
squaremo, line 54: I don't think we should be using a build tag to segregate tests (see my comment in #387), so `--build-tags integration` should be redundant here.

squaremo (commented):
Ohh yes, linting is good, thank you -- and following up linter warnings is also welcome!
Let's resolve #387 and come back to this one.

.github/workflows/pull-request.yml
stevesloka: 😱 Yeah that doesn't make any sense.

PR #376, opened 2022-11-23
Skip dependency installation option in API

squaremo:
Addresses #374 by adding a field to the Stack API, telling the controller to skip the dependency installation step.

PR #389, opened 2022-12-21
flux-source: Add skipInstallDependencies flag

### LINKED ISSUE #374: Abilities to skip install dependencies
## Hello!

## Issue details

It would be great to have flag to skip installing dependencies. In my case, I use poetry to build and package code with all dependencies as OCI artifact. When Pulumi operator executed, it failed to install dependencies because requirements.txt is not in the OCI artifact. So in this case, it does not need to install dependencies because they are already included in artifact file.

### Affected area/feature

Pulumi Operator

stevesloka:
### Proposed changes

Adds a skipInstallDependencies flag to the flux source integration which disables the automatic dependency installation.

### Related issues (optional)

Fixes #374

### DIFF
--- a/deploy/crds/pulumi.com_stacks.yaml
+++ b/deploy/crds/pulumi.com_stacks.yaml
@@ -189,6 +189,10 @@ spec:
                       project (i.e., containing Pulumi.yaml) of interest, within the
                       fetched source.
                     type: string
+                  skipInstallDependencies:
+                    description: SkipInstallDependencies can be set to true to skip
+                      automatically installing dependencies.
+                    type: boolean
                   sourceRef:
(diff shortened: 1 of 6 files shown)


squaremo:
Is this considered complete or is it a draft?

squaremo:
There's a [draft PR already](https://github.com/pulumi/pulumi-kubernetes-operator/pull/376) for this feature, by the way. Roo and I got as far as changing the API and outlining a test for the new behaviour, so (if my interpretation above is accurate) it's a little further along than this one.

stevesloka:
Didn't realize there was already work in progress @squaremo, I can close this PR out.

PR #407, opened 2023-02-01
Use stable working directory paths

### LINKED ISSUE #381: `go build` use of disk space leads to pod eviction
While inspecting a running operator pod, we found 15GB of files under /tmp/.cache, created by `go build` (i.e., by a Go-based Pulumi project used by s Stack object). In the [canonical deployment](https://github.com/pulumi/pulumi-kubernetes-operator/blob/master/deploy/yaml/operator.yaml) of the operator, /tmp has an ephemeral volume mounted; the kubelet logs indicated that it had evicted the operator pod based on the node running low on ephemeral storage space:

```
Message:          The node was low on resource: ephemeral-storage. Container pulumi-kubernetes-operator was using 82316Ki, which exceeds its request of 0. [...]
```

In general we can't stop Pulumi programs using arbitrary amounts of disk space, and that is an _operational_ issue, to be dealt with by careful configuration of pod disk limits, monitoring, and coping with pod evictions gracefully. But: we may be able to stop Go stockpiling files, and remove that as a cause of evictions.

squaremo:
Previously, the operator asked the OS to create a fresh tmp directory for each stack run. This is a reasonable enough design, since it means builds are always from a fresh clone, can't interfere with one another, and can be safely deleted afterwards. But it also meant that the go build cache would treat every source file as distinct source code, and grow its build cache accordingly -- to the tune of tens or hundreds of megabytes a build.

This commit changes the working directory to a stable path (still under /tmp), named for the stack object. Since each stack object is uniquely named and is processed serially, there should be no collisions. (There's also no security benefit to having random directory names, since a program or user will ill intent can always just poke around in the filesystem to find what they want).

This also centralises the creation of the directory, which removes a bit of repetition.

Fixes #381.

---
Further notes:

This will have a doubly beneficial effect on Go projects, since the go build cache will work as intended -- so, no unbounded disk use, but also much faster builds. I checked this by making a project which vendors its libraries and using it in a stack:

 - `pulumi new kubernetes-go`, then `go mod vendor; git add .; git commit -m "Initial revision"`
 - create a stack referring to that git repo
 - get a shell into the operator pod, and `watch du -sh /tmp/.cache/go-build`

When using stable paths, that cache grows to 320MB the first time it processes the stack, then stays there. Without this change, the cache directory grew by (just about) 320MB each time the stack was processed.


rquitales (commented):
LGTM, just a minor comment.

pkg/controller/stack/stack_controller.go
```diff
@@ -1027,7 +1044,30 @@ func (sess *reconcileStackSession) lookupPulumiAccessToken(ctx context.Context)
 	return "", false
 }
 
-func (sess *reconcileStackSession) SetupWorkdirFromGitSource(ctx context.Context, gitAuth *auto.GitAuth, source *shared.GitSource) (_revision string, retErr error) {
+// Make a working directory for building the given stack. These are stable paths, so that (for one
+// thing) the go build cache does not treat new clones of the same repo as distinct files. Since a
+// stack is processed by at most one thread at a time, and stacks have unique qualified names, and
+// the directory is expected to be removed after processing, this won't cause collisions; but, we
+// check anyway, treating the existence of the build directory as a crude lock.
+func makeWorkingDir(s *pulumiv1.Stack) (_path string, _err error) {
```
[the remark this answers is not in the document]
squaremo: Not really; I occasionally have the impulse to name return values that have ground types, just for the documentation.

squaremo:
Thanks for having a look @rquitales 🙇

squaremo:
I built a dev image for this: `docker.io/squaremo/pulumi-kubernetes-operator:581425f`
(digest sha256:f28997f94f5d3f5e9e4fc7ad0e5e4f1b41fcdeff1b3c5a17b49c28da13a6483b)

pkg/controller/stack/stack_controller.go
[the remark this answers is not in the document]
squaremo: Do you think I should remove the names?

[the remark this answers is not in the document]
rquitales: I'll defer the choice to you. Normally, I wouldn't have named returns, especially if the var names are discarded or the function is simple, but consistency within the project would be more important.

test/suite_test.go
```diff
@@ -163,7 +163,11 @@ func resetWaitForStack() {
 
 // internalWaitForStackState refetches the given stack, until its .lastUpdated.state field matches
 // the one given. NB it fetches into the pointer given, so mutates the struct it's pointing at.
-func internalWaitForStackState(stack *pulumiv1.Stack, state shared.StackUpdateStateMessage) {
+func internalWaitForStackState(stack *pulumiv1.Stack, state shared.StackUpdateStateMessage, maybeTimeout ...string) {
```
rquitales, line 166: nit: `maybeTimeout` wasn't self-explanatory as a variable name to me initially. Perhaps `optionalTimeout` might be a better name?

test/stable_build_test.go
```diff
@@ -0,0 +1,107 @@
...
+	When("a go stack is run multiple times", func() {
+		var (
+			beforeSize string
+		)
+
+		checkCacheSize := func() string {
+			out, err := exec.Command("go", "env", "GOCACHE").Output()
+			ExpectWithOffset(1, err).ToNot(HaveOccurred())
+			cachedir := string(out)
+			cachedir = strings.TrimSpace(cachedir)
```
rquitales, line 77: Can you also add an assert here to ensure that we're using the tempdir cache directory?

test/stable_build_test.go
```diff
@@ -0,0 +1,107 @@
...
+			GinkgoWriter.Println("Before:", beforeSize)
+			// make sure the cache is actually used!
+			Expect(beforeSize).NotTo(HavePrefix("0\t"))
+		})
+
+		It("doesn't grow the Go build cache", func() {
+			// run it again and check that we get the same result from du -sh <cache>
+			stack.Spec.ResyncFrequencySeconds = 30 // just to provoke reprocessing
+			resetWaitForStack()
+			Expect(k8sClient.Update(context.TODO(), stack)).To(Succeed())
```
rquitales, line 100: Would it be beneficial to run this in a loop for a few more times for sanity checking as well? Or add another go build step and validation steps so we can verify the cache doesn't grow with more than just 2 runs?

test/suite_test.go
squaremo: Sure. Different traditions :-)

test/stable_build_test.go
squaremo: My reasoning here is that the cache growing _at all_ after the first build is enough evidence. If you try it at home, when a fresh path is constructed for each build you'll see the cache grow on the second and subsequent builds, by the same increment -- so we do have the control case (that could be encoded into a test, if you think it would bolster confidence that stable paths makes a difference).

rquitales: Spoke offline. Since each run takes ~90s, we'll opt for faster test runs so no change is necessary.

rquitales (approved):

PR #415, opened 2023-02-17
Put GITHUB_TOKEN in env to mitigate API throttling

squaremo:
GitHub has quite strict limits on API requests if you are an unauthenticated user, which the tests can run into since they download Pulumi plugins from GitHub. In the context of GitHub Actions, there is an access token available, which will give elevated API quotas. Actions can use it by default, but it needs to be made available in the environment for it to be in effect in tests.

Hence: add an environment entry for GITHUB_TOKEN to any workflow that runs the tests.


kpitzen (approved):

squaremo:
Thanks Kyle :star:

PR #416, opened 2023-02-17
Less fickle check for cache growth

squaremo:
In CI, the test that checks whether building at a stable path works with the go build cache occasionally fails. The test checks the output of `du` (disk usage) before and after, and fails (when it fails) with the difference of a small number of blocks.

Since the test is to check whether new objects are added to the cache, it's not necessary to check the size. On the assumption that the change in block usage is incidental, this commit changes that test so it compares the (sorted, recursive) _list_ of file names before and after. New files would indicate new objects being introduced, and the test will fail (and if the change to stable build paths is reverted, that's what happens).

### DIFF
--- a/test/stable_build_test.go
+++ b/test/stable_build_test.go
@@ -67,8 +67,8 @@ var _ = Describe("go build caching", func() {
 
 	When("a go stack is run multiple times", func() {
 		var (
-			beforeSize string
-			cachedir   string
+			beforeFiles string
+			cachedir    string
 		)
 
 		getCachedir := func() string {
@@ -77,13 +77,14 @@ var _ = Describe("go build caching", func() {
 			return strings.TrimSpace(string(out))
 		}
 
-		checkCacheSize := func() string {
+		checkCacheFiles := func() string {
 			d := getCachedir()
 			GinkgoWriter.Println("Cache:", d)
-			// there's just one source file in the project, so the difference between the cache
-			// working and it not working will be pretty small. `-k` measures in 1KiB blocks, which
-			// should be fine enough.
-			out, err := exec.Command("du", "-s", "-k", d).Output()
+			// there's just one source file in the project, so the size difference between the cache
+			// working and it not working will be pretty small. However, there certainly should not
+			// be _new_ files after building again; so, listing all the files, should catch cache
+			// growth.
+			out, err := exec.Command("ls", "-aR", d).Output()
 			ExpectWithOffset(1, err).ToNot(HaveOccurred())
 			s := string(out)
 			return s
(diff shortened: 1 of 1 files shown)


pkg/controller/stack/stack_controller.go
```diff
@@ -1050,6 +1050,7 @@ func (sess *reconcileStackSession) lookupPulumiAccessToken(ctx context.Context)
 // the directory is expected to be removed after processing, this won't cause collisions; but, we
 // check anyway, treating the existence of the build directory as a crude lock.
 func makeWorkingDir(s *pulumiv1.Stack) (_path string, _err error) {
+	//	d, _ := os.MkdirTemp("", "pulumi-build")
```
squaremo, line 1053: I can take this out; I left it here because it was convenient for checking whether the test passed when the build directory was _not_ stable (and the go build cache thereby got new files in it).

kpitzen: Agreed - this should probably be omitted now.

kpitzen (commented):
Just a general question here - does a (even small) change in block usage of cache files _not_ indicate that cache growth is occurring?  Would we expect it to grow by large amounts each iteration?

test/stable_build_test.go
```diff
@@ -77,13 +77,14 @@ var _ = Describe("go build caching", func() {
...
 			GinkgoWriter.Println("Cache:", d)
-			// there's just one source file in the project, so the difference between the cache
-			// working and it not working will be pretty small. `-k` measures in 1KiB blocks, which
-			// should be fine enough.
-			out, err := exec.Command("du", "-s", "-k", d).Output()
+			// there's just one source file in the project, so the size difference between the cache
+			// working and it not working will be pretty small. However, there certainly should not
+			// be _new_ files after building again; so, listing all the files, should catch cache
+			// growth.
+			out, err := exec.Command("ls", "-aR", d).Output()
```
rquitales, line 87: Happy with shelling out to `ls`. But also prefer not to shell out when possible, so using`filepath.Walk` or `os.ReadDir`. I'll defer the choice to you.

rquitales:
> Just a general question here - does a (even small) change in block usage of cache files _not_ indicate that cache growth is occurring? Would we expect it to grow by large amounts each iteration?

I believe the original issue is that cache is rebuilt every time when we do not have a stable path for reconciling our stacks. So our go build cache size (and number of files) would be expected to grow linearly the more times we reconcile. This updates the test to ensure that the number of files in the cache doesn't grow, ie relatively constant. We made the assumption that we could check cache growth using storage size as a measure, but minor differences in the cache file sizes after reconciles makes this comparison harder, so using distinct counts of files is a better metric.

rquitales (approved):
Minor comments that don't really need to be addressed. LGTM

squaremo:
> does a (even small) change in block usage of cache files _not_ indicate that cache growth is occurring?

Ramon's answer is absolutely on point. I am not sure why there's a difference in block use. It's always a small number of blocks difference to an existing file. It happens routinely in CI, but not nearly as much when the tests are run locally. It would seem that the compiler is recompiling a file, without its cache key changing. But I care much less about individual cache entries changing a little than about new cache entries -- the latter would indicate that stable paths do not fix the problem, and anyway be orders of magnitude difference.

test/stable_build_test.go
squaremo: Yeah, I have the same preference, in general. I am OK with this instance because it's more or less a one-liner, and that invocation should be stable across OSes (ubiquitous command, and standard flags).

PR #426, opened 2023-03-23
Use more accurate Kubernetes capitalization and naming

rquitales:


deploy/crds/pulumi.com_stacks.yaml
```diff
@@ -868,13 +868,13 @@ spec:
                 type: boolean
               destroyOnFinalize:
                 description: (optional) DestroyOnFinalize can be set to true to destroy
-                  the stack completely upon deletion of the CRD.
+                  the stack completely upon deletion of the Stack custom resource.
```
rquitales, line 871: For review: changed to the more accurate term of custom resource (instead of CustomResourceDefinition).

thomas11 (approved):

PR #443, opened 2023-05-04
Refresh dynamic credentials in provider state using stack refresh

rquitales:


deploy/crds/pulumi.com_stacks.yaml
```diff
@@ -549,6 +549,37 @@ spec:
                   preferred first, then personal access token, and finally basic auth
```
rquitales, line 549: Note: autogenerated CRD from Golang structs.

docs/stacks.md
```diff
@@ -189,6 +189,13 @@ StackSpec defines the desired state of Pulumi Stack being managed by this operat
           (optional) GitAuthSecret is the the name of a Secret containing an authentication option for the git repository. There are 3 different authentication options: * Personal access token * SSH private key (and it's optional password) * Basic auth username and password Only one authentication mode will be considered if more than one option is specified, with ssh private key/password preferred first, then personal access token, and finally basic auth credentials. Deprecated. Use GitAuth instead.<br/>
```
rquitales, line 189: Note: autogenerated

pkg/apis/pulumi/shared/stack_types.go
```diff
@@ -5,6 +5,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
rquitales, line 5: Adding the two new API fields for our Stack CR: `Targets` and `Prerequisite`. This is how our users will use this added feature.

pkg/apis/pulumi/v1/stack_types.go
```diff
@@ -18,6 +18,9 @@ type StackStatus struct {
 	// ObservedGeneration records the value of .meta.generation at the point the controller last processed this object
```
rquitales, line 18: Adding additional status fields for our controller to keep track of reconcile progression.

pkg/controller/stack/stack_controller.go
```diff
@@ -138,16 +144,10 @@ func add(mgr manager.Manager, r *ReconcileStack) error {
...
-	//  - https://github.com/operator-framework/operator-sdk/issues/2795
-	//  - https://github.com/kubernetes-sigs/kubebuilder/issues/1103
-	//  - https://github.com/kubernetes-sigs/controller-runtime/pull/553
-	//  - https://book-v1.book.kubebuilder.io/basics/status_subresource.html
-	// Set up predicates.
+	// Filter for update events where an object's metadata.generation is changed (no spec change!),
+	// or the "force reconcile" annotation is used (and not marked as handled).
 	predicates := []predicate.Predicate{
-		predicate.Or(predicate.GenerationChangedPredicate{}, libpredicate.NoGenerationPredicate{}),
+		predicate.Or(predicate.GenerationChangedPredicate{}, ReconcileRequestedPredicate{}),
```
rquitales, line 150: Predicates give us finer control of the Kubernetes events the controller sees, and allows it to control which resources need to be reconciled.

pkg/controller/stack/stack_controller.go
```diff
@@ -19,7 +19,6 @@ import (
 	"time"
```
rquitales, line 19: NB: most logic added in this file determines how/when a Stack CR should be reconciled.

test/stack_status_test.go
```diff
@@ -3,7 +3,6 @@ package tests
 import (
```
rquitales, line 3: Code was moved into `suite_test.go` since the `makeFixtureIntoRepo` function is used by most tests.

test/stale_state_test.go
```diff
@@ -0,0 +1,257 @@
+// Copyright 2022, Pulumi Corporation.  All rights reserved.
```
rquitales, line 1: Testing the flow of this feature. Here we create a RabbitMQ stack/container, and another stack that interacts with RabbitMQ. We then go through our expectations that it:

1. Works expected normally if nothing is stale
2. Fails if the container attempting to ping RabbitMQ has old credentials in State
3. Succeeds to interact with RabbitMQ if we create a new Stack CR to specifically refresh the stale secret
4. Succeeds if we define the new stack as a prerequisite refresh Stack, and the parent stack is enqued.

test/suite_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"context"
```
rquitales, line 6: Moving code around, no new function added.

rquitales, line 6: Moving code around, no new function added.

pkg/controller/stack/stack_controller.go
```diff
@@ -285,6 +317,71 @@ func add(mgr manager.Manager, r *ReconcileStack) error {
 	return nil
 }
 
+// isRequirementSatisfied checks the given readiness requirement against the given stack, and
+// returns nil if the requirement is satisfied, and an error otherwise. The requirement can be nil
+// itself, in which case the prerequisite is only that the stack succeeded onm its last run.
```
kpitzen, line 322:
```suggestion
// itself, in which case the prerequisite is only that the stack succeeded on its last run.
```
nit

pkg/controller/stack/stack_controller.go
```diff
@@ -285,6 +317,71 @@ func add(mgr manager.Manager, r *ReconcileStack) error {
...
+// This protocol is the same mechanism used by many Flux controllers, as explained at
+// https://pkg.go.dev/github.com/fluxcd/pkg/runtime/predicates#ReconcileRequestedPredicate and
+// related documentation.
+type ReconcileRequestedPredicate struct {
+	predicate.Funcs
+}
+
+func getReconcileRequestAnnotation(obj client.Object) (string, bool) {
+	r, ok := obj.GetAnnotations()[shared.ReconcileRequestAnnotation]
+	return r, ok
```
kpitzen, line 367: Nit: Do we need to destructure this return?

kpitzen, line 367: Do we need to destructure this return? Can we just return `obj.GetAnnotations...`? This call will need to be treated as a tuple, either way, yeah?

kpitzen:
If I'm understanding this PR correctly, it seems like we're targeting this:
> Support a mechanism to allow stack authors to selectively target specific URNs for update before running refresh. This is essentially the equivalent of running pulumi up -t <provider> as in the CLI workaround mentioned above. With this the credential refresh code could live within the pulumi program and leverage the pulumi sdk instead of in a hand-coded script.
From https://github.com/pulumi/pulumi-kubernetes-operator/issues/299 - is that correct?

pkg/controller/stack/stack_controller.go
rquitales: Since we want to test whether the key exists in the map or not, we actually need to do a 2 value assignment (ie. the comma ok idiom). Simply returning without assigning will only result in one value being returned, the value of the map key.

Example:
https://go.dev/play/p/x9NarquVj5X

rquitales:
> If I'm understanding this PR correctly, it seems like we're targeting this:
> 
> > Support a mechanism to allow stack authors to selectively target specific URNs for update before running refresh. This is essentially the equivalent of running pulumi up -t  as in the CLI workaround mentioned above. With this the credential refresh code could live within the pulumi program and leverage the pulumi sdk instead of in a hand-coded script.
> 
> From #299 - is that correct?

Yes, that is correct.

pkg/controller/stack/stack_controller.go
rquitales: Done.

pkg/controller/stack/stack_controller.go
kpitzen: Golang isn't a real language, is it? ;)

kpitzen (approved):

PR #445, opened 2023-05-08
Rquitales/arm images

rquitales:
Adds building ARM64 images. Requires https://github.com/pulumi/pulumi-docker-containers/pull/140 to be merged with based ARM images built and distributed.


blampe:
Plan is to strip down the operator image & spawn jobs using multi-arch images as part of https://github.com/pulumi/pulumi-kubernetes-operator/issues/515.

PR #448, opened 2023-05-10
Serialize specs to be ordered to prevent flakeyness

rquitales:
Some specs need to be run in a certain order, or they would fail. This is a quick fix to ensure we do not have  flakey tests, but comes at a performance hit. We should look into decoupling these specs if possible.

We also create resources (eg. ConfigMaps) with the same name and namespace. This will cause race conditions and dead locks. This PR renames some resources to prevent this, but we should have proper namespace isolation for these tests to run in parallel.

This is a quick fix for now to unblock the v1.12.0 release.


test/testdata/test-program.yaml
```diff
@@ -12,7 +12,7 @@ resources:
                 annotations:
                     pulumi.com/patchForce: "true"
                 name:
-                    test-configmap
+                    test-configmap-valid-test
```
rquitales, line 15: Prevent deletion race condition, as another spec was previously creating a CM with the same name as well.

rquitales:
Superseded in a later PR. Note that this might be of use in the future as the later commits introduces spinning up kind clusters for testing.

PR #480, opened 2023-08-22
feat: add cluster role and fluxcd crd rules to role

### LINKED ISSUE #479: Allow adding role and rolebinding for flux source access
## Hello!
Currently the helm chart doesn't include any way to allow the operator to access flux source. It would be good to include it.

### Affected area/feature

```[tasklist]
### Tasks
```

dirien:
### Proposed changes

Hi,

this PR adds optional cluster role support plus adding the FluxCD CR to the (non-cluster) role

### Related issues (optional)

Closes #479


mikhailshilkov:
@dirien Are you looking into the test failure?

PR #490, opened 2023-09-09
Use a per-stack PULUMI_HOME directory

### LINKED ISSUE #483: Use different ACCESS_TOKENS for different Stacks
### What happened?

I set up 2 Stacks with 2 different Stack Resources. I used EnvRefs to specify my Pulumi Access Token:
```yaml
  envRefs:
    PULUMI_ACCESS_TOKEN:
      type: Literal
      literal:
        value: <redacted>
```
I also tried setting the token via SecretRef
```yaml
    PULUMI_ACCESS_TOKEN:
      type: Secret
      secret:
        name: tokens
        key: accessToken
```
If i got only one Stack or if i can access all Stacks  with the same token, it works perfectly fine. If i have two Stacks that need two different Access Tokens, it does not work anymore.

It always uses the Token from the first Stack it runs (after restarting the pod), and keeps using it for all Stacks it processes. It errors with "Organisation not found" and if i have a look at the "wrong" organisation, it tried to create the Stack in the wrong org because it havent found it there. Even though i specified the correct Tokens in each Stack Resource.

### Expected Behavior

Use the correct Token.

### Steps to reproduce

Create Resources as described.

### Output of `pulumi about`

I run version 1.13.0.

### Additional context

_No response_

### Contributing

EronWright:
### Proposed changes
Closes #483 

This PR seeks to isolate the credentials associated with a given `Stack`, to solve the problem of credentials leaking across stacks. Some underlying details here:
1. Pulumi CLI stores login credentials in PULUMI_HOME (e.g. `~/.pulumi/credentials.json`).
2. A side-effect of using `PULUMI_ACCESS_TOKEN` is that the CLI login credentials are set.
4. Pulumi CLI prefers the persisted login credentials to `PULUMI_ACCESS_TOKEN`.

This PR takes the conservative approach of encapsulating the PULUMI_HOME into a per-stack working directory, as opposed to reusing `~/.pulumi` across stacks. The working directory is retained across reconciliation passes, and cleaned up during stack finalization. Note that the _workspace_ directory is erased at the end of each reconciliation pass, as is the current behavior.

This PR does NOT solve the (lack of) mutability of `PULUMI_ACCESS_TOKEN` across stack updates.

_Note that this PR contains some commits (related to hacking on the operator) that will be moved to a separate PR._

### Technical Details
Relevant terminology used within the controller codebase:
- **root directory** - a temporary directory for each stack, retained until finalization
- **home directory** - the `PULUMI_HOME` directory, located within the stack's root directory
- **workspace directory** - the Pulumi workspace directory containing the program and stack configuration.

The current behavior of the operator is to erase the workspace directory on each reconciliation pass, e.g. to ensure a clean git checkout. This PR retains this behavior while keeping the home directory across passes, e.g. to reuse the providers. 

### Related issues (optional)
- https://github.com/pulumi/pulumi/issues/13919


pkg/controller/stack/stack_controller.go
```diff
@@ -1220,27 +1227,74 @@ func (sess *reconcileStackSession) lookupPulumiAccessToken(ctx context.Context)
 	return "", false
 }
 
-// Make a working directory for building the given stack. These are stable paths, so that (for one
+// Make a root directory for the given stack, containing the home and workspace directories.
+func (sess *reconcileStackSession) MakeRootDir(ns, name string) (_path string, _err error) {
```
iwahbe, line 1231: If you don't plan to use the named return values, we generally don't have any.
```suggestion
func (sess *reconcileStackSession) MakeRootDir(ns, name string) (string, error) {
```

If you really want to use named return values, we don't need to include the `_` in names.

pkg/controller/stack/stack_controller.go
```diff
@@ -988,6 +991,10 @@ func (sess *reconcileStackSession) finalizeStack(ctx context.Context) error {
 			return err
 		}
 	}
+
+	// Delete the root directory for this stack.
+	sess.cleanupRootDir()
```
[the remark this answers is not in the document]
EronWright: That's right. Also, errors in the local cleanup stage are ignored simply because that was the existing behavior.

pkg/controller/stack/stack_controller.go
EronWright: I agree, that was due to copy-paste of the `MakeWorkspaceDir` func.

test/stack_isolation_test.go
```diff
@@ -0,0 +1,125 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const buildDirectoryPrefix = "pulumi-working"
+
+var _ = Describe("Stack isolation", func() {
+	var (
+		tmpDir             string
+		gitDir, backendDir string
```
rquitales, line 25: nit: We could combine these lines:
```suggestion
		tmpDir, gitDir, backendDir string
```

test/stack_isolation_test.go
```diff
@@ -0,0 +1,125 @@
...
+		Expect(k8sClient.Create(context.TODO(), stack)).To(Succeed())
+		waitForStackSuccess(stack)
+		Expect(exists(filepath.Join(rootDir, "workspace"))).To(BeFalse())
+	})
+
+	finalizationTest := func() {
+		It("should purge the root directory", func() {
+			rootDir := getRootDir(stack)
+			Expect(k8sClient.Create(context.TODO(), stack)).To(Succeed())
+			waitForStackSuccess(stack)
```
rquitales, line 106: Can we also check that the rootDir exists here as an extra check.

test/stack_isolation_test.go
```diff
@@ -0,0 +1,125 @@
...
+		Expect(k8sClient.Create(context.TODO(), stack)).To(Succeed())
+		waitForStackSuccess(stack)
+		Expect(exists(rootDir)).To(BeTrue())
+		Expect(exists(filepath.Join(rootDir, ".pulumi"))).To(BeTrue())
+	})
+
+	It("should purge the workspace dir", func() {
+		rootDir := getRootDir(stack)
+		Expect(k8sClient.Create(context.TODO(), stack)).To(Succeed())
+		waitForStackSuccess(stack)
```
rquitales, line 98: It would be nice to also somehow include a check that the workspace folder existing when the stack is being reconciled to validate that we are indeed using a workspace folder.

pkg/controller/stack/stack_controller.go
```diff
@@ -1220,27 +1227,74 @@ func (sess *reconcileStackSession) lookupPulumiAccessToken(ctx context.Context)
...
+	homeDir := sess.getPulumiHome()
+	if err := os.MkdirAll(homeDir, 0700); err != nil {
+		return "", fmt.Errorf("error creating .pulumi dir: %w", err)
+	}
+	return rootDir, nil
+}
+
+// cleanupRootDir cleans the root directory that contains the Pulumi home and workspace directories.
+func (sess *reconcileStackSession) cleanupRootDir() {
+	if sess.rootDir != "" {
```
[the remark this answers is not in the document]
EronWright: Agreed, I too prefer line-of-sight code style.

test/stack_isolation_test.go
EronWright: Sorry I don't see a decent way to do that.

rquitales: I was thinking of just spawning a goroutine that polls the existence of the workspace folder. Something like:
```
It("should purge the workspace dir", func() {
	rootDir := getRootDir(stack)
	Expect(k8sClient.Create(context.TODO(), stack)).To(Succeed())
	// Spy on the root directory in separate goroutine to make sure the workspace dir is created.
	var workspaceCreated bool
	done := make(chan struct{})
	go func() {
		for {
			select {
			case <-time.After(2 * time.Second):
				if exists(filepath.Join(rootDir, "workspace")) {
					workspaceCreated = true
					return
				}
			case <-done:
				return
			}
		}
	}()
	waitForStackSuccess(stack)
	close(done) // Clean-up spawned goroutine.
	Expect(workspaceCreated).To(BeTrue())
	Expect(exists(filepath.Join(rootDir, "workspace"))).To(BeFalse())
})
```

EronWright: I'm not thrilled at the lack of determinism, I think I'll try to make a hook for testing purposes.

rquitales: Let's expand on this in a follow-up PR.

rquitales (approved):

PR #491, opened 2023-09-10
vscode project files

EronWright:
This PR contributes a basic vscode project configuration for ease of use.  The gitignore is derived from [here](https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/Global/VisualStudioCode.gitignore) (same as [here](https://www.toptal.com/developers/gitignore/api/go,vim,emacs,visualstudiocode)).

### Proposed changes

The project files do the following:
1. Configure the default Build Task (⇧⌘B) to run `make build`.
2. Configure the default Test Task to run `make test`.
3. Install a Launch Configuration to run `cmd/manager` in the debugger.
4. Recommend some vscode extensions when the project is opened (Go, Kubernetes).

~Also, introduces an (undocumented) environment variable to skip leader election, for dev purposes: `DISABLE_LEADER_ELECTION=true`.~

https://github.com/pulumi/pulumi-kubernetes-operator/assets/1775518/278d3248-7d5d-47b5-a8d5-b2420d5e43f0


cmd/manager/main.go
```diff
@@ -119,12 +120,14 @@ func main() {
 
 	log.Info("Graceful shutdown", "timeout", gracefulShutdownTimeout)
 
+	disableElection := cmdutil.IsTruthy(os.Getenv("DISABLE_LEADER_ELECTION"))
```
iwahbe, line 123: This should probably be a different PR.

.gitignore
```diff
@@ -73,9 +76,15 @@ Session.vim
 tags
 ### VisualStudioCode ###
 .vscode/*
-.history
-# End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
-pulumi-kubernetes-operator
-### IntelliJ
+!.vscode/settings.json
```
iwahbe, line 79: Knowing nothing about VSCode configuration, why do we exclude `.vscode/*` and then add back certain files. Do we expect that contributors will have other non-checked in files in `.vscode`?

.vscode/tasks.json
```diff
@@ -0,0 +1,26 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"type": "shell",
+			"label": "make: build",
+			"command": "make build",
+			"group": {
+                "kind": "build",
```
iwahbe, line 9: Let's fix the formatting here.

.gitignore
danielrbradley: Agreed. Perhaps we can just remove the .vscode ignore?

EronWright: Looking around on the Web, the approach shown here is the norm (see links in description). Assumedly there are good reasons, can we just trust the crowd here?

thomas11: These are all distinct file names in all my .vscode folders, via a `find | sort | uniq` command.

```
configurationCache.log
cspell-azd-dictionary.txt
cspell-dictionary.txt
cspell-github-user-aliases.txt
cspell-notice.txt
cspell-templates.txt
cspell.global.yaml
cspell.yaml
dryrun.log
extensions.json
launch.json
settings.json
targets.log
tasks.json
```

I don't even know what `cspell` is, and we don't want the `log` files. So the approach here makes sense, based on my n=1 sample. I guess we could also allow all .json and .yaml files?

EronWright: I would advocate for us to stick to the well-known approach in our first iteration, which is to whitelist a few files but ignore everything else. Folks may raise a follow-up PR as we learn about other stuff.

.vscode/tasks.json
EronWright: Tempting to add an `.editorconfig` to control the tab chars.

iwahbe (approved):
LGTM

thomas11 (approved):

PR #548, opened 2024-02-06
Fix `nodeSelector`, `affinity`, `tolerations` in the Helm chart

v-yarotsky:
### Proposed changes

`nodeSelector`, `affinity`, and `tolerations` are currently ignored unless `initContainers` is set to a non-empty value, due to a misplaced `if` in the template.
This change fixes the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ CHANGELOG
 =========
 
 ## HEAD (unreleased)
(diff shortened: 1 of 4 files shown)


rquitales (approved):
Changes looks good to me. I've also bumped the versions for our charts. @dirien can you do a quick review as well to make sure this flow is correct?

dirien:
@v-yarotsky, nice catch! That was a big L from my side! 😱

dirien (approved):
LGTM!

Issue #544, opened 2024-01-23
helm install command doesn't work in docs


SnoozeFreddo:
### What happened?

Can't install the operator with the given instructions from the documentation.

### Example

helm repo add pulumi-kubernetes-operator https://pulumi.github.io/pulumi-kubernetes-operator/

### Output of `pulumi about`

Error: looks like "https://pulumi.github.io/pulumi-kubernetes-operator/" is not a valid chart repository or cannot be reached: failed to fetch https://pulumi.github.io/pulumi-kubernetes-operator/index.yaml : 403 Forbidden

### Additional context

It's because its a chart source but not a [chart repository](https://helm.sh/docs/topics/chart_repository/)

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mjeffryes:
Thanks for filing this ticket @SnoozeFreddo; I wasn't able to find the documentation that suggests running `helm repo add pulumi-kubernetes-operator https://pulumi.github.io/pulumi-kubernetes-operator/` Would you mind pointing me to where you saw that? Alternatively, if you'd like to contribute any improvements to the documentation for this project, we'd be happy to review them!

mjeffryes:
Update: I did manage to find the offending instructions here: https://github.com/pulumi/pulumi-kubernetes-operator/blob/5fd9efc38a2420048e5fb69cc0b444fd6a8a252b/deploy/helm/pulumi-operator/README.md?plain=1#L22

PR #561, opened 2024-03-12
Update README.md to remove outdated helm instructions

mjeffryes:
We apparently only support installing via the oci artifact. (https://github.com/pulumi/pulumi-kubernetes-operator/issues/544#issuecomment-1980777127) 

Fixes https://github.com/pulumi/pulumi-kubernetes-operator/issues/544

### DIFF
--- a/.github/workflows/chart-lint-and-test.yml
+++ b/.github/workflows/chart-lint-and-test.yml
@@ -3,6 +3,8 @@ on:
   pull_request:
     paths:
       - deploy/helm/**
+      - "!deploy/helm/pulumi-operator/README.md"
+      - "!deploy/helm/pulumi-operator/README.md.gotmpl"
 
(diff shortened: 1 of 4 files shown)


mjeffryes:
The linting check error seems to just be that it wants a chart version bump, but I don't think that makes sense b/c this is just a readme change.

rquitales (approved):
@mjeffryes FYI reworked this PR to:

1. Replicate the readme changes you made into the readme template, since the publish steps uses the template file for generating a readme
2. Updated the title of the readme - I noticed it's using the name of another project
3. Skip chart linting and publish if only the readme is updated

I'll let you do a final look over and have you merge it.

mjeffryes:
Totally missed that this was a template! Thanks for cleaning up after me @rquitales!

PR #575, opened 2024-04-25
Support structured config

ljtfreitas:
### Context

Currently structured configs [are not supported properly](https://github.com/pulumi/pulumi-kubernetes-operator/issues/258) by Pulumi k8s operator. 

As Pulumi programs can be configured using YAML files, it looks fair to be able to configure a structured, complex config using the Stack CRD as well.

### Proposed changes

The proposed approach is to introduce a new field called `configRefs`, rather than change the current `config` one.

`configRefs` has a format close to other fields like `envRefs` and `secretsRefs`, supporting the same `env` and `filesystem` options as well (`secret` is not supported because just use `secretsRefs` looks better), and two new options:

- ConfigLiteralRef
- ConfigMapRef

#### ConfigLiteralRef

ConfigLiteralRef refers to both a simple or possibly structured, nested value in YAML format, in the same way that people would put in regular `Pulumi.<stack>.yaml` files:

```yaml
configRefs:
   stack-config:
     type: Literal
     value:
        my-config:
          my-nested-field:
             field: "value"
```

The `value` content will be handled as a map of any type of value; so just simple key-pairs in YAML format are supported as well:

```yaml
configRefs:
   stack-config:
     type: Literal
     value:
       field-a: value-a
       field-b: value-b
```

#### ConfigMapRef

ConfigMapRef refers to a Kubernetes ConfigMap. The implementation assumes that the content from the ConfigMap's key is a structured, nested content in YAML format. The idea is to enable people to just put their configs from `Pulumi.<stack>.yaml` files in a ConfigMap, and the operator will be able to handle it.

So given a ConfigMap called `my-stack-config`,  like that:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-stack-config
data:
  Pulumi.my-stack.yaml: |
    my-config:
      my-nested-field:
          field: "value"
```

It could be consumed as a config to a Pulumi stack like that:

```yaml
configRefs:
   stack-config:
     type: ConfigMap
     configmap:
       name: my-stack-config
       key: Pulumi.my-stack.yaml
```

#### Configs merge strategy

All configs resolved from `configRefs` fields will be merged with the ones from the `config` field. `configRefs` resolved key-values will override `config` ones with the same key.

#### Configs `set` strategy

All configs will be set using the `path` flag. So the `config` field will start to support nested values as well.

It's especially useful for some complex fields from providers. An example:

```yaml
config:
   aws:assumeRole.roleArn: arn/my-aws-arn
```

Passing the value above with the `--path` flag, it will be configured like that in the result stack config:

```yaml
aws:assumeRole:
  roleArn: arn/my-aws-arn
```

And simple key-value pairs keep working also.

### Related issues (optional)

https://github.com/pulumi/pulumi-kubernetes-operator/issues/258
https://github.com/pulumi/pulumi-kubernetes-operator/pull/516


ljtfreitas:
cc @rquitales @EronWright

blampe:
This is awesome, thank you for the contribution! One initial thought -- still need to take a deeper look.

> The idea is to enable people to just put their configs from Pulumi.<stack>.yaml files in a ConfigMap, and the operator will be able to handle it.

This is a great idea. What do you think about using Pulumi's native marshalers? That would give you a little validation on top, and it might also make your flattening logic unnecessary (although I'm not 100%). The [encoding](https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3@v3.113.3/go/common/encoding) package should do the trick, and you would unmarshal into a [ProjectStack](https://pkg.go.dev/github.com/pulumi/pulumi/sdk/go/common/workspace#ProjectStack).

ljtfreitas:
@blampe Thanks. About your suggestion maybe I haven't expressed myself properly. My initial idea, and that's the implementation is actually doing, is to put in a ConfigMap just the `config` node from a `Pulumi.<stack>.yaml` file, not the whole ProjectStack. `Pulumi.<stack>.yaml` can contain fields like `secretsprovider`, as the struct that you mentioned have, but it was not my idea to be able to configure them through a ConfigMap.

However, it looks like a good idea to me 🤔 ; so we could move the whole stack file (`Pulumi.<stack>.yaml`) to a ConfigMap and we even wouldn't have to configure things like `secretsprovider` or `backend` in the CRD 🤔 ; we could just unmarshal the ConfigMap key's content as a ProjectStack. Not sure if it's related with the specific topic of structured config but looks good to me.

blampe:
> @blampe Thanks. About your suggestion maybe I haven't expressed myself properly. My initial idea, and that's the implementation is actually doing, is to put in a ConfigMap just the `config` node from a `Pulumi.<stack>.yaml` file, not the whole ProjectStack. `Pulumi.<stack>.yaml` can contain fields like `secretsprovider`, as the struct that you mentioned have, but it was not my idea to be able to configure them through a ConfigMap.
> 
> However, it looks like a good idea to me 🤔 ; so we could move the whole stack file (`Pulumi.<stack>.yaml`) to a ConfigMap and we even wouldn't have to configure things like `secretsprovider` or `backend` in the CRD 🤔 ; we could just unmarshal the ConfigMap key's content as a ProjectStack. Not sure if it's related with the specific topic of structured config but looks good to me.

Ah, I had misunderstood what all was in the ConfigMap, I didn't realize it was just the config key. In that case you could marshal the [config.Map](https://pkg.go.dev/github.com/pulumi/pulumi/sdk@v1.14.1/go/common/resource/config#Map) instead of the whole ProjectStack as I had suggested earlier. I didn't mean to increase the scope of your change!

ljtfreitas:
@blampe Don't worry, maybe I haven't expressed myself properly in the PR description (English is not my first language so sometimes talking about complex stuff is just...complex 😅). About your suggestion, I'm trying to refactor the code to use `config.Map` but I'm struggling to read the content. When I try to unmarshal I see this error:

```
could not parse <my-config> as a configuration key (configuration keys should be of the form `<namespace>:<name>`
```

Looking at the code that deserializes a YAML to a ProjectStack instance (in the Automation API code), config keys are namespaced (with the project name) before the config.Map's unmarshal 😕 ...and I can't use the same function because it's private. So config keys must be namespaced in the ConfigMap.

Not sure if we need to enforce that here because this behavior will be applied anyway during the config-set phase (and speaking for myself I don't namespace my configs because of this)

ljtfreitas:
hey friends, any news here? we really need this feature 

@blampe @rquitales

blampe:
Sorry for the delay @ljtfreitas, I plan on playing around with this tomorrow!

test/stack_controller_test.go
```diff
@@ -205,6 +207,386 @@ var _ = Describe("Stack Controller", func() {
...
+					Fail("Unable to set WORD environment variable.")
+				}
+
+				// Define the stack spec
+				localSpec := shared.StackSpec{
+					Backend: fmt.Sprintf("file://%s", backendDir),
+					Stack:   stackName,
+					GitSource: &shared.GitSource{
+						ProjectRepo: baseDir,
+						RepoDir:     "test/testdata/config-refs",
```
blampe, line 295: Could we just `test/testdata/structured-config-refs` for all of these? This is testing we can read the configs, seems worth also confirming the structure was handled correctly.

pkg/controller/stack/stack_controller.go
```diff
@@ -1085,6 +1086,75 @@ func (sess *reconcileStackSession) SetEnvRefsForWorkspace(ctx context.Context, w
...
+				}
+				// assumes the whole configmaps's data is the config content; try to read as a conventional stack yaml config
+				var configMapContent map[string]any
+				if err := yaml.Unmarshal([]byte(config.Data[configMapRef.Key]), &configMapContent); err != nil {
+					return nil, fmt.Errorf("Failed to read the ConfigMap content as a stack YAML config. Namespace=%s Name=%s: %w", configMapRef.Namespace, configMapRef.Name, err)
+				}
+				structuredConfig := StructuredConfig(configMapContent).Flatten()
+				allConfigs = append(allConfigs, structuredConfig...)
+			}
+		case shared.ConfigResourceSelectorStructured:
```
blampe, line 1109: I'm a little confused about the need for both "structured" and "literal".

The legacy `resolveResourceRef` logic doesn't do any special handling for "literal", but we have an opportunity to be more flexible here. 

In other words, it's odd to have two ways to express the same thing:
```
configRefs:
   stack-config:
     type: Literal
     value: foo
--
configRefs:
   stack-config:
     type: Structured
     value: foo
```

But this is net-new:
```
configRefs:
   stack-config:
     type: Structured
     value:
        my-config:
          my-nested-field:
             field: "value"
```

It would seem reasonable IMO to basically remove structured and just change how we handle literals (basically replacing this block with  `case ResourceSelectorLiteral`).

blampe (commented):
Apologies again for the delay. I was able to spend some hands-on time with this and have a much better feel for things. 

The biggest issue I see is that I think we'll want `ConfigRefs` to be an array. Let me know if you think that sounds reasonable.

I'm still playing around with some of the config merging logic locally, in case there's some room to simplify things.

ljtfreitas:
thank you @blampe, don't worry, don't need to apologize (i hope that I have not sounded like a jerk or something 😅 also), i will check the comments 👀

test/testdata/structured-config-refs/go.mod
```diff
@@ -0,0 +1,90 @@
+module structured-config-refs
```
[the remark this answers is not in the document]
ljtfreitas: I guess I was a bit biased here in fact 😅, I wrote the example/test program in Go because I wanted to be sure that we could deserialize a structured, complex config in the way that we usually do in Go

deploy/crds/pulumi.com_stacks.yaml
```diff
@@ -80,6 +80,105 @@ spec:
...
+                        of selector. Must be one of: Env, FS, Secret, ConfigMap, Structured,
+                        Literal'
+                      type: string
+                  required:
+                  - type
+                  type: object
+                description: (optional) ConfigRefs is the configuration for this stack,
+                  which can be specified through ConfigRef. If this is omitted, configuration
+                  is assumed to be checked in and taken from the source repository.
+                  If present, ConfigRefs values will be merged with values passed
```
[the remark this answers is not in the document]
ljtfreitas: but ConfigRefs will not overwrite Configs, right? in fact both fields will be merged

I guess the description could be improved to explain the `ConfigRefs` will override `Configs` with the same key name

deploy/crds/pulumi.com_stacks.yaml
```diff
@@ -80,6 +80,105 @@ spec:
...
                   is assumed to be checked in and taken from the source repository.
                 type: object
+              configRefs:
+                additionalProperties:
+                  description: ConfigRef identifies a resource from which config information
+                    can be loaded. Environment variables, files on the filesystem,
+                    Kubernetes Secrets, ConfigMaps, structured and config literal
+                    values strings are currently supported.
+                  properties:
+                    configmap:
```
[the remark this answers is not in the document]
ljtfreitas: Interesting. It could be an array, but do we really need this? 🤔 `ConfigRefs` would have the same semantics as `EnvRefs` field.

In EnvRefs, one could specify more than one secret, or more than one env var, with the same key name and the order will not be guaranteed; but this would look like a misconfiguration to me (using the same secret key twice), the behavior seems reasonable.

also, the `ConfigRefs` spec above supports to specify the same ref type more than once, like:

```
configRefs:
   first-secret:
      type: Secret
      secret:
        name: my-secret
        key: key-name
   second-secret:
      type: Secret
      secret:
        name: my-another-secret
        key: key-name
```

pkg/controller/stack/stack_controller.go
ljtfreitas: I thought about this option first but the reason why I've preferred to introduce the "structured" type was don't change the "literal" behavior, because it's already used by `EnvRefs` and `SecretRefs` fields, and for these use cases the "literal" option looks good enough (specially for env var case)

For these use cases just passing a literal, simple string value makes sense, but a structured, complex value does not, I think...so I didn't want to change that, and introduced the new "structured" option (I used this word because Pulumi docs also use it to [reference about complex configs](https://www.pulumi.com/docs/concepts/config/#structured-configuration). So the main reason was backward compatibility.

but you're right, it sounds a bit strange. maybe we could handle "literal" for configs as you suggested and also keep the legacy "literal" behavior for the other `Refs` fields. WDYT?

pkg/controller/stack/stack_controller.go
```diff
@@ -1556,20 +1626,46 @@ func (sess *reconcileStackSession) InstallProjectDependencies(ctx context.Contex
...
+	configValuesFromRefs, err := sess.resolveConfigRefs(ctx)
+	if err != nil {
+		return fmt.Errorf("Fail reading ConfigRef values: %w", err)
 	}
+	configValues = append(configValues, configValuesFromRefs...)
+
+	// config values should be ordered by key to avoid problems like try to set 'c[1]=value1' before 'c[0]=value2'
+	sort.Slice(configValues, func(i, j int) bool {
+		return configValues[i].Key < configValues[j].Key
+	})
```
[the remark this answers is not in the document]
ljtfreitas: Is the second example a valid syntax? as a json?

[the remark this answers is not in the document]
blampe: I just doubled checked -- these are two ways of describing the same thing:

```
  stack:deepobject:
    nums:
      - 1
  stack:flatobject: '{"nums": [1]}
```

It doesn't look like `stack:object.nums[0]: 1` is valid.

I think it would be totally fine to say this is an edge case we don't handle, given that it's unlikely to come up.

deploy/crds/pulumi.com_stacks.yaml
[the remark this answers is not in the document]
ljtfreitas: Ok, it makes sense. I will change it 👍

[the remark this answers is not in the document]
ljtfreitas: Thinking about the best way to do it 🤔 ...maybe an "array of maps"?

```yaml
configRefs:
   - base:
        type: Structured 
        value:
           my-config: base-value
   - override:
        type: Structured
        value:
           my-config: override-value
```

Or an object with a "name" field (or "key" or something)?

```yaml
configRefs:
   - name: base
      type: Structured 
      value:
         my-config: base-value
   - name: override
      type: Structured
      value:
         my-config: override-value
```

I prefer the first one but any thoughts?

test/stack_controller_test.go
ljtfreitas: Actually this test is trying to read a file, and in a similar way that I implemented with `Literal`, I used the same existing behavior to `FileRef` (the operator already use it to load envvars and secrets from files), which means that just loads a single value. So this test couldn't use `test/testdata/structured-config-refs`

Do you believe that would be worth implementing a `file` behavior specifically to load configs? So we could support structured values as well.

deploy/crds/pulumi.com_stacks.yaml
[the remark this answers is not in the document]
ljtfreitas: @blampe just added a new test case for this scenario

ljtfreitas:
hi friends 👋 , any news here? cc @blampe

miguelslemos:
Hi @blampe, this feature is crucial for us, could you take a look when you get time, please?

pkg/controller/stack/stack_config_test.go
```diff
@@ -0,0 +1,219 @@
...
+			"aws:assumeRole": map[string]any{
+				"roleArn":     "my-role-arn",
+				"sessionName": "my-session-name",
+			},
+			"aws:defaultTags": map[string]any{
+				"tags": map[string]any{
+					"my-tag": "tag-value",
+				},
+			},
+		}
```
blampe, line 56: For what it's worth, I think if you add another key like `"aws:assumeRole.roleArn": "other-role-arn"` this will fail non-deterministically.

pkg/controller/stack/stack_controller.go
```diff
@@ -1085,6 +1086,85 @@ func (sess *reconcileStackSession) SetEnvRefsForWorkspace(ctx context.Context, w
 	return nil
 }
 
+func (sess *reconcileStackSession) resolveConfigRefs(ctx context.Context) ([]ConfigKeyValue, error) {
+	allConfigs := make([]ConfigKeyValue, 0)
+	for _, configRefMap := range sess.stack.ConfigRefs {
+		for k, ref := range configRefMap {
```
blampe, line 1092: I'm not clear on why this is a `[]map[string]ConfigRef` instead of just a `[]ConfigRef`. The key only seems to be used for secret refs (which mirrors `SecretRefs`?), and for the default case (do we need that?).

pkg/controller/stack/stack_controller.go
```diff
@@ -1085,6 +1086,85 @@ func (sess *reconcileStackSession) SetEnvRefsForWorkspace(ctx context.Context, w
...
+				}
+				allConfigs = append(allConfigs, ConfigKeyValue{
+					Key: k,
+					Value: auto.ConfigValue{
+						Value:  secretValue,
+						Secret: true,
+					},
+				})
+			default:
+				// try to resolve as a ResourceRef
```
blampe, line 1142: Do we need this? `resolveResourceRef` doesn't support structure so it seems odd to default to it. An error would be reasonable IMO.

blampe (commented):
Sorry, have been OOO. I'm still a little uncertain about the type of `ConfigRefs`, and I worry we're duplicating some of `SecretRefs`. It would be nice to test merging behavior when keys collide.

> Hi @blampe, this feature is crucial for us, could you take a look when you get time, please?

You should be able to use a custom image if this is blocking you.

pkg/controller/stack/stack_config_test.go
ljtfreitas: Could be but, sorry, is it a real problem? For a stack configured like this,

```yaml
aws:assumeRole:
   roleArn: "my-role-arn"
aws:assumeRole.roleArn: "other-role-arn"
```

I wouldn't expect any deterministic behavior here, it would look like a configuration mistake to me.

pkg/controller/stack/stack_controller.go
ljtfreitas: It was a design decision, that I've mentioned here: https://github.com/pulumi/pulumi-kubernetes-operator/pull/575#discussion_r1603536059

ljtfreitas: I've chosen the `array of maps` option to be similar to the `envRefs` and `secretRefs` fields, where the key is the env/secret key name. 

But in fact, for `configRefs`, the key is not used in the code for `literal` and `configMap` options. But I don't know what could be a better option 😕

pkg/controller/stack/stack_controller.go
ljtfreitas: It doesn't, but `resolveResourceRef` supports all other cases like `env`, `file`, etc. These options are supported in `configRefs` too. 

Maybe it does not make sense and `configRefs` should support just structured/config map options?

pkg/controller/stack/stack_controller.go
ljtfreitas: For literal values it's used as the config-key name:

```yaml
configRefs:
   - sample:
        type: Literal
        value: just-a-value
```

it will be resolved as a config like that:

```
sample: just-a-value
```

pkg/controller/stack/stack_controller.go
```diff
@@ -1085,6 +1086,85 @@ func (sess *reconcileStackSession) SetEnvRefsForWorkspace(ctx context.Context, w
...
+					// ConfigLiteralRef handles both simple and structured values as json, flattening all keys to build a list of Pulumi key:value configs
+					structuredConfig, err := NewStructuredConfigFromJSON(k, literalRef.Value)
+					if err != nil {
+						return nil, fmt.Errorf("Failed to unmarshall %s as a structured config: %w", k, err)
+					}
+					configs := structuredConfig.Flatten()
+					allConfigs = append(allConfigs, configs...)
+				}
+			// Secret should be handled here as well because auto.ConfigValue should be marked as Secret:true
+			case shared.ConfigResourceSelectorType(shared.ResourceSelectorSecret):
```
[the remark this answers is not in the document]
ljtfreitas: You wouldn't, it doesn't make sense. I will remove it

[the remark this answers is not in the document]
ljtfreitas: My original idea was to support the same options from `ResourceRef` but I'm wondering about that...maybe should we support just literal/structured values and ConfigMap for configs?

ljtfreitas:
hey @blampe , I've made some changes and updated the PR description; I've removed the `structured` concept and started to handle just as `literal` values, both simple or complex ones.

pkg/controller/stack/stack_config.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+type StructuredConfig map[string]any
+
+type ConfigKeyValue struct {
+	Key   string
+	Value auto.ConfigValue
+}
+
+func NewStructuredConfigFromJSON(key string, rawValue apiextensionsv1.JSON) (*StructuredConfig, error) {
```
blampe, line 19: Don't we want to accept YAML?

pkg/controller/stack/stack_config_test.go
blampe: I know it may seem weird, but it's possible so we need to consider it. Hyrum's law.

It would be very helpful to see some more tests around what happens when multiple config maps are provided with overlapping keys like this.

pkg/controller/stack/stack_controller.go
```diff
@@ -1085,6 +1086,64 @@ func (sess *reconcileStackSession) SetEnvRefsForWorkspace(ctx context.Context, w
...
+						return nil, fmt.Errorf("Failed to read the ConfigMap content as a stack YAML config. Namespace=%s Name=%s: %w", configMapRef.Namespace, configMapRef.Name, err)
+					}
+					structuredConfig := StructuredConfig(configMapContent).Flatten()
+					allConfigs = append(allConfigs, structuredConfig...)
+				}
+			case shared.ConfigResourceSelectorLiteral:
+				literalRef := configRef.Literal
+				if literalRef != nil {
+					// ConfigLiteralRef handles both simple and structured values as json, flattening all keys to build a list of Pulumi key:value configs
+					structuredConfig, err := NewStructuredConfigFromJSON(k, literalRef.Value)
```
blampe, line 1114: Already mentioned above, but interpreting literal as an object eliminates the need for `k` while still allowing structure.

```
literal:
  value:
    aws:region: us-west-2
    foo:bar:
      - baz
```

pkg/controller/stack/stack_controller.go
blampe: Yes, we should keep the new API surface area as small as possible. Unless we have a good reason to include them now, we can omit `env`, `file` etc. and add them later if needed. We don't need a key `k` for the literal case which leaves this default case as the only reason for it. Not needed IMO.

blampe (commented):
I think we can still simplify the API a bit. Let's try to scope this down to only the things we know we need, which it sounds like is config maps and literals?

Other things that will help get this over the line:
* Blocker: Resolve non-determinism by making ConfigRefs an array.
* Blocker: Tests around edge cases with merging multiple refs/literals.
* Optional: We could limit the change to an alpha API, to leave room for changes before we add it to v1.

mikhailshilkov:
@ljtfreitas Are you still planning to move this PR forward and address the comments?

ljtfreitas:
Sorry for the late. No, I'm going to close the PR

Unfortunally the operator v2 has the same issue 😕 .

PR #580, opened 2024-05-10
Bump pulumi to 3.114 in published image

blampe:


rquitales (approved):

rquitales:
Actually, can we bump to v3.115.2 to match #579?

Issue #611, opened 2024-07-24
Implement proto server methods


EronWright:
A subset of the auto api for purposes of the PKOv2 initiative (deploy operations).  Other methods would be added later as follow-ups (stack import/export).

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

Issue #606, opened 2024-07-20
gRPC layer for Automation API


EronWright:
This issue tracks the implementation tasks for a new gRPC layer for the Automation API; see [Proposal](https://docs.google.com/document/d/1hHOgs9uDsvAbj89D-b4WjVG9GnoJ7_cy5FKGOeSAfRg/edit?usp=sharing).

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

PR #624, opened 2024-08-02
[pkov2] agent RPC server

### LINKED ISSUE #610: Develop the grpc proto specification
Finalize the subset of auto API methods to be exposed via gRPC, based on the needs of the PKO project.

- `WhoAmI`
- `Info`
- `Refresh`
- `Preview`
- `Up`
- `Destroy`

If stack selection happens implicitly during the operation, there's probably no need for any explicit stack management commands.

Use streaming responses as appropriate, e.g. for deployment operations where an event stream is available.

EronWright:
### Proposed changes

**Epic Link**: https://github.com/pulumi/pulumi-kubernetes-operator/issues/606

**Demo Video Link**: https://pulumi.slack.com/archives/C07DQSV84DC/p1722636430008649

Implements an agent consisting of two commands:

- `init` - fetches a flux source into the given directory, intended for use in an init container.
- `serve` - starts an RPC server providing an automation API to perform stack updates over the given workspace.

### Overview
The RPC server assumes that the project source code has been checked out to a local working directory, called the "workspace" directory.  This generally corresponds to a sub-directory within a git repository, e.g. [examples/random-yaml](https://github.com/pulumi/examples/tree/master/random-yaml).

At startup, the server opens the workspace using [auto.NewLocalWorkspace](https://github.com/pulumi/pulumi/blob/5651750bb254f73da5ef0fa503818c5a38755ea8/sdk/go/auto/local_workspace.go#L848). All RPC operations are applied to this same workspace, usually one-at-a-time. Some operations cause state changes, e.g. stack selection, that may affect subsequent operations.  Some operations produce `PreconditionFailed` if a stack hasn't been selected.

At startup, the server optionally runs `pulumi install` to install dependencies and plugins for the project, based on https://github.com/pulumi/pulumi/pull/16782. Note that PKOv1 has some code to install plugins, now believed to be obsolete (see [discussion](https://github.com/pulumi/pulumi/pull/16782#issuecomment-2259286745)).

The supported operations are:

- `WhoAmI` - returns current user info.
- `Install` - runs `pulumi install` in the workspace.
- `SelectStack` - select (and optionally create) a stack, for use in subsequent operations.
- `Info` - a summary of the current stack.
- `SetAllConfig` - set multiple configuration values on the current stack, based on literals, environment variables, and file references.  It is expected that the server's pod would have ConfigMaps and Secrets mounted accordingly.
- `Preview` runs the preview operation for the current stack.
- `Up` runs the up operation for the current stack.
- `Destroy` runs the destroy operation for the current stack.
- `Refresh` runs the refresh operation for the current stack.

The deployment operations have streaming responses, consisting of a series of engine events and a final result.

The agent uses zap for logging, because it supports structured logging, implements `io.Writer` to capture Pulumi console output, and integrates well with grpc-go.

### Follow-ups

- [x] Write RPC server tests
- [ ] Rename 'init' to 'fetch' for clarity
- [ ] lock the workspace during an operation? Or rely on locking within the Pulumi CLI?

### Related issues (optional)

Closes #610 #611


agent/go.mod
```diff
@@ -0,0 +1,114 @@
+module github.com/pulumi/pulumi-kubernetes-operator/agent
```
blampe, line 1: Structure nits:

* We should put the new code under a v2 subdirectory.
* We should try as hard as possible to keep one `go.mod` at the root of the repo, unless there's a legitimate version incompatibility that's non-trivial to resolve between v1 and v2. (Is there?)
* If we absolutely need a new module, it should be under v2 so the agent and server both belong to it since they're versioned together.

Unnecessary modules are a big source of accidental complexity and lead to all sorts of headaches and gotchas down the line ([example](https://github.com/pulumi/pulumi-service/pull/20778)). Trust me on this one :)

agent/go.mod
```diff
@@ -0,0 +1,114 @@
...
+	github.com/spf13/cobra v1.8.1
+	go.uber.org/zap v1.26.0
+	google.golang.org/grpc v1.65.0
+	google.golang.org/protobuf v1.34.2
+	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
+)
+
+// Replace digest lib to master to gather access to BLAKE3.
+// xref: https://github.com/opencontainers/go-digest/pull/66
+replace github.com/opencontainers/go-digest => github.com/opencontainers/go-digest v1.0.1-0.20220411205349-bde1400a84be
```
blampe, line 20: Any background on this? I'm not familiar with it, just curious what it gets us.

agent/main.go
```diff
@@ -0,0 +1,22 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package main
+
+import "github.com/pulumi/pulumi-kubernetes-operator/agent/cmd"
+
+func main() {
+	cmd.Execute()
+}
```
blampe, line 22: Structure nit: this is probably more at home under `cmd` since that's where binaries typically live. Something like this
```
v2
├── cmd
│   ├── agent
│   │   ├── init.go
│   │   ├── main.go
│   │   ├── root.go
│   │   └── serve.go
│   └── operator
│       └── main.go
├── go.mod (optional -- "module github.com/pulumi/pulumi-kubernetes-operator/v2")
├── internal
│   ├── operataor
│   └── server
└── proto
```

Note also the absence of `pkg` -- "a package to hold other packages" doesn't really serve a purpose and is considered an anti-pattern. On the other hand, `internal` can effectively accomplish the same while also allowing you to not worry about public APIs until they actually matter (which can be never).

agent/cmd/serve.go
```diff
@@ -0,0 +1,190 @@
...
+	Use:   "serve",
+	Short: "Serve the agent RPC service",
+	Long: `Start the agent gRPC server.
+`,
+	PreRun: func(cmd *cobra.Command, args []string) {
+		rpcLogger = zap.L().Named("rpc")
+		grpc_zap.ReplaceGrpcLoggerV2(rpcLogger)
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := cmd.Context()
```
blampe, line 59: Not something you need to do now, but suggest eventually making these cobra functions pass-throughs to your server package. There's some non-trivial setup that's worth testing eventually, and it'll be easier to do that without the cobra overhead.

agent/cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+)
+
+var verbose bool
+
+// a command-specific logger
+var log *zap.SugaredLogger
```
blampe, line 28: Structured logging ❤️

agent/pkg/proto/agent.proto
```diff
@@ -0,0 +1,272 @@
...
+
+package agent;
+
+service AutomationService {
+
+  /**
+   * WhoAmI returns detailed information about the currently logged-in Pulumi
+   * identity.
+   */
+  rpc WhoAmI(WhoAmIRequest) returns (WhoAmIResult) {}
```
blampe, line 16: Let's call these responses instead of results -- I think it's more idiomatic for proto.
```suggestion
  rpc WhoAmI(WhoAmIRequest) returns (WhoAmIResponse) {}
```

agent/pkg/proto/agent.proto
```diff
@@ -0,0 +1,272 @@
+syntax = "proto3";
+
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/pulumi/pulumi-kubernetes-operator/agent/pkg/proto";
+
+package agent;
```
blampe, line 8: I think this should be `package proto` to match its subdirectory but I could be wrong. It might be more expected to have this live in the top-level agent dir with generated code down here.

agent/pkg/proto/agent.proto
```diff
@@ -0,0 +1,272 @@
...
+  /**
+   * Up creates or updates the resources in a stack, returning the actual
+   * changes.
+   */
+  rpc Up(UpRequest) returns (stream UpStream) {}
+
+  /**
+   * Destroy deletes all resources in a stack.
+   */
+  rpc Destroy(DestroyRequest) returns (stream DestroyStream) {}
```
blampe, line 59: I think the stream types can also be called responses since the streaming-ness is obvious from the context.

agent/pkg/proto/agent.proto
```diff
@@ -0,0 +1,272 @@
...
+message WhoAmIResult {
+  string user = 1;
+  repeated string organizations = 2;
+  string url = 3;
+}
+
+message SelectStackRequest {
+  string stackName = 1;
+  optional bool create = 2;
+  optional string secrets_provider = 3;
```
blampe, line 75:
```suggestion
  string stack_name = 1;
  optional bool create = 2;
  optional string secrets_provider = 3;
```

agent/pkg/proto/agent.proto
```diff
@@ -0,0 +1,272 @@
...
+  optional bool path =
+      1; // Parse the keys as paths in a map or list rather than raw strings
+  map<string, ConfigValue> config =
+      2; // a map of ConfigValue used by Pulumi programs
+}
+
+message ConfigValue {
+  oneof v {
+    google.protobuf.Value value = 1;
+    ConfigValueFrom valueFrom = 2;
```
blampe, line 103:
```suggestion
    ConfigValueFrom value_from = 2;
```

agent/pkg/proto/agent.proto
```diff
@@ -0,0 +1,272 @@
+syntax = "proto3";
+
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/pulumi/pulumi-kubernetes-operator/agent/pkg/proto";
+
+package agent;
+
+service AutomationService {
```
blampe, line 10: No required args 👍 👍

agent/pkg/proto/agent.proto
```diff
@@ -0,0 +1,272 @@
...
+  string name = 1;
+  google.protobuf.Timestamp last_update = 2;
+  bool update_in_progress = 3;
+  optional int32 resource_count = 4;
+  optional string url = 5;
+}
+
+message SetAllConfigRequest {
+  optional bool path =
+      1; // Parse the keys as paths in a map or list rather than raw strings
```
blampe, line 95: Am I right in thinking this basically resolves the structure config problem by letting you mount any config map you want into the pod and use that as your stack config directly? Does that mean SetAllConfig would only get invoked once per operation?

agent/pkg/server/server_test.go
```diff
@@ -0,0 +1,809 @@
...
+			name:       "new stack",
+			projectDir: "./testdata/simple",
+			opts:       &server.Options{StackName: "new"},
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			g := gomega.NewWithT(t)
```
blampe, line 64: ❤️

blampe (approved):
This is an awesome start! Most of my feedback is nits or structural suggestions that don't need to block.

agent/pkg/proto/agent.proto
EronWright: This `SetAllConfig` RPC method closely mirrors the underlying auto api, that takes a `path` bool to indicate that the keys in the provided map are "paths" as opposed to literal config keys. What's special here is that the RPC method is able to resolve "valueFrom" refs that draw values from environment variables and local files.

The Stack controller calls `SetAllConfig` today after extracting values from configmaps and secrets.  In the v2 design, the controller doesn't see the values. The configmaps/secrets are mounted onto the workspace pod (as envs and/or files) and then read via these "valueFrom" refs.

agent/go.mod
EronWright: Thanks for the reminder, I added a link to where this originated (fluxcd):
https://github.com/fluxcd/pkg/blob/968f1907a8d27b8a13ee40fab88ae7dbdb553539/http/fetch/go.mod#L10-L12

agent/version/version.go
```diff
@@ -0,0 +1,3 @@
+package version
+
+var Version string = "v0.0.1"
```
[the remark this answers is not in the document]
EronWright: Thanks, I tried to use ReadBuildInfo but [apparently](https://github.com/golang/go/issues/50603) it works only with `go install` not `go build`, and so I opted for the status quo.

agent/go.mod
EronWright: I plan to tackle the larger structural stuff in a follow-up.

Issue #652, opened 2024-09-06
Setup an "in-cluster" kubeconfig within the workspace pod


EronWright:
PKOv1 makes an effort to create a kubeconfig within the operator's own container at startup (see [`setupInClusterKubeconfig`](https://github.com/pulumi/pulumi-kubernetes-operator/blob/fb0a7f1246cb3f99e2a73f8460ca498f8132579e/pkg/controller/stack/stack_controller.go#L1886)). The rationale is, I believe, to facilitate the use of the pulumi-kubernetes provider in your program.

Since the program now runs in the workspace pod, we may need to do similarly, either by adding a default kubeconfig to the pu/pu image, or by creating a kubeconfig during workspace initialization. Of course, this is simply an "in-cluster" connection to the API server using the pod's service account.

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

EronWright:
@julienp for your awareness

blampe:
Can write some code in the agent to generate an in-cluster config.

rquitales:
This work is not required for the Pulumi Kubernetes provider to work with the cluster that the pod is running in. Our default config loading rules (from client-go) in p-k will already attempt to use the inCluster credentials if it can't find a valid kubeconfig. I have manually validated this by deploying a `pulumi/pulumi-go` pod, which does not have any Kubernetes tooling installed, and it is able to create k8s resources from a Pulumi program within the same cluster the pod is running in.

Ref to our config loading rules: https://github.com/pulumi/pulumi-kubernetes/blob/74673fdf488bdf3513cd76d6f6d62ae0184faabc/provider/pkg/provider/provider.go#L779-L785

PR #658, opened 2024-09-10
Git source

### LINKED ISSUE #655: Implement the Git source
The new stack controller doesn't actually support the `git` spec yet; it works only with Flux.  To complete the git support, we need:

- [x] Develop a way to query the git remote to resolve the HEAD of the given branch, ideally without cloning the repo (as is done today).
- [x] Implement support for git auth options, incl. in the workspace pod where the clone operation takes place.

blampe:
Open questions:
* I'm currently passing auth information via env vars to the workspace init container. Do we want to provide them a different way?
* The current API is allows auth to be specified via literals, env vars or on-disk files but these don't make sense in an agent/server model. I'm inclined to say we should only support secret refs for these fields, but this is technically a breaking change for anyone currently specifying literals/vars/files. Otherwise I think we'll need to mirror the flexible types in the workspace and add some logic to the agent to resolve them at runtime (might need https://github.com/pulumi/pulumi-kubernetes-operator/issues/652?). What do you guys think?

Summary of changes so far:
* Fetch latest git commit using an in-memory object store.
* Agent's init now uses automation API to clone the repo. Seems like a natural place to dogfood.
* ~~Removed code for `gitAuthSecret` since it was marked deprecated in v1.~~
* Exposed a "shallow" option to the stack API for controlling shallowness. We don't currently perform shallow clones so this is backwards compatible.

Fixes #655


operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -1016,6 +1002,35 @@ func (sess *StackReconcilerSession) resolveResourceRefAsConfigItem(ctx context.C
 	}
 }
 
+// resolveResourceRef reads a referenced object and returns its value as a string.
+func (sess *StackReconcilerSession) resolveResourceRef(ctx context.Context, ref *shared.ResourceRef) (string, error) {
```
EronWright, line 1006: Hopefully we can remove this method, and defer resolution of the value to the workspace pod.  In other words the operator shouldn't see the resolved value.

operator/internal/controller/pulumi/git.go
```diff
@@ -18,318 +18,240 @@ import (
...
-	// 				"Namespace", sess.namespace, "Stack.GitAuthSecret", sess.stack.GitAuthSecret)
-	// 			return nil, err
-	// 		}
+	// Otherwise fetch the most recent commit for the ref (branch) we care
+	// about.
+	auth, err := gs.authMethod()
+	if err != nil {
+		return "", fmt.Errorf("getting auth methodL: %w", err)
+	}
+	err = gs.remote.FetchContext(ctx, &git.FetchOptions{
```
EronWright, line 103: Did it prove impractical to avoid fetching the bits?

EronWright (commented):
Looking great, and thanks for using `GIT_` vars.

On the topic of SSH known hosts, what's the minimum we need to do? Keep in mind that the user can customize the pu/pu docker image.

I too encountered the "breaking change" you mentioned, when porting the stack configuration code. The "local file" or "local env variable" reference makes little sense now, and would it be interpreted as a reference to the operator pod or to the workspace pod?  I think we should flag it as a migration issue, and eventually add validation logic to reject it.

operator/internal/controller/pulumi/git.go
blampe: Updated this to essentially perform an ls-remote and then grep the remote refs. Tests went from ~1.5s to ~1.0s when run against p-examples.

operator/internal/controller/pulumi/stack_controller.go
blampe: Discussed offline - operator needs to resolve these in order to poll the remote for latest commit.

blampe:
> On the topic of SSH known hosts, what's the minimum we need to do? Keep in mind that the user can customize the pu/pu docker image.

I removed this TODO. Instead of accepting whatever the host's key is currently, we should encourage the use to customize the workspace with a custom mount containing the expected keys.

agent/cmd/init.go
```diff
@@ -76,61 +71,38 @@ For Flux sources:
...
-}
+		repo := auto.GitRepo{
+			URL: GitURL,
+			//ProjectPath: ,
+			CommitHash: GitRevision,
+			Auth:       auth,
+			Shallow:    os.Getenv("GIT_SHALLOW") == "true",
+		}
+
+		_, err = auto.NewLocalWorkspace(ctx, auto.Repo(repo))
```
EronWright, line 98: Expecting the code to be checked out the the TargetDir, which is `/share/source`.

operator/go.mod
```diff
@@ -3,7 +3,6 @@ module github.com/pulumi/pulumi-kubernetes-operator/operator
 go 1.22.5
 
 require (
-	github.com/fluxcd/pkg/http/fetch v0.11.0
```
EronWright, line 6: Odd to see this package be removed.

operator/internal/controller/auto/workspace_controller.go
```diff
@@ -434,6 +441,61 @@ func newStatefulSet(ctx context.Context, w *autov1alpha1.Workspace, source *sour
...
+				Value: source.Git.Dir,
+			},
+		}
+		if source.Git.Shallow {
+			env = append(env, corev1.EnvVar{
+				Name:  "GIT_SHALLOW",
+				Value: "true",
+			})
+		}
+		// TODO: These are only available to the init container -- Should they
```
EronWright, line 464: I think no with the rationale that the vars should remain a private contract between the operator and the agent.

operator/internal/controller/pulumi/git.go
```diff
@@ -18,318 +18,276 @@ import (
...
-	// 	}
-
-	// 	if u.Scheme == "ssh" && gitAuth.SSHPrivateKey == "" {
-	// 		return gitAuth, fmt.Errorf("a private key must be provided for SSH")
-	// 	}
-
-	// 	return gitAuth, nil
+// Source represents a source of commits.
+type Source interface {
+	CurrentCommit(context.Context) (string, error)
```
EronWright, line 40: Is `Source` an interface that the Flux stuff should also use?

operator/internal/controller/pulumi/git.go
```diff
@@ -18,318 +18,276 @@ import (
...
 
 	git "github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
-	"github.com/go-git/go-git/v5/plumbing/protocol/packp/capability"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/go-git/go-git/v5/plumbing/transport/ssh"
+	"github.com/go-git/go-git/v5/storage/memory"
+	"github.com/pulumi/pulumi-kubernetes-operator/operator/api/auto/v1alpha1"
```
EronWright, line 29: Please give this package the alias `autov1alpha1`.

operator/internal/controller/pulumi/git.go
```diff
@@ -18,318 +18,276 @@ import (
...
 		}
+	}
 
-		// Firstly we will try to check that an SSH Private Key Path has been specified
-		if authDetails.SSHPrivateKeyPath != "" {
-			publicKeys, err := ssh.NewPublicKeysFromFile("git", repoArgs.Auth.SSHPrivateKeyPath, repoArgs.Auth.Password)
-			if err != nil {
-				return "", fmt.Errorf("unable to use SSH Private Key Path: %w", err)
-			}
+	return "", fmt.Errorf("no commits found for ref %q", gs.ref)
```
EronWright, line 117: When this (or similar spec problem) happens, does the status transition to Stalled as it should?

operator/internal/controller/pulumi/git.go
```diff
@@ -18,318 +18,276 @@ import (
...
 	}
 
-	// *Repository.Clone() will do appropriate fetching given a branch name. We must deal with
-	// different varieties, since people have been advised to use these as a workaround while only
-	// "refs/heads/<default>" worked.
-	//
-	// If a reference name is not supplied, then .Clone will fetch all refs (and all objects
-	// referenced by those), and checking out a commit later will work as expected.
-	if repoArgs.Branch != "" {
-		refName := plumbing.ReferenceName(repoArgs.Branch)
```
EronWright, line 245: There's some sophistication here in decoding the reference.  Should some of this logic survive?

operator/internal/controller/pulumi/session_test.go
```diff
@@ -16,7 +26,6 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/auto"
```
EronWright, line 26: Odd to see `auto` imported here.  Would have expected the autov1alpha package.

EronWright (changes requested):

operator/go.mod
blampe: It's only needed by the agent. Overhead from juggling 2 mod files ;)

```
~/src/pulumi-kubernetes-operator/operator
❯ go mod why github.com/fluxcd/pkg/http/fetch
# github.com/fluxcd/pkg/http/fetch
(main module does not need package github.com/fluxcd/pkg/http/fetch)

~/src/pulumi-kubernetes-operator/agent
❯ go mod why github.com/fluxcd/pkg/http/fetch
# github.com/fluxcd/pkg/http/fetch
github.com/pulumi/pulumi-kubernetes-operator/agent/cmd
github.com/fluxcd/pkg/http/fetch
```

operator/internal/controller/auto/workspace_controller.go
blampe: Agree!

operator/internal/controller/pulumi/git.go
blampe: Only if it would be helpful for the purpose of mocking GitHub/Flux in unit tests -- otherwise we don't need it.

operator/internal/controller/pulumi/git.go
blampe: Yes, with SourceUnavailable.

operator/internal/controller/pulumi/git.go
blampe: Not worth keeping -- this is more to do with the actual repo checkout, which we're now offloading to the automation API. For example `refs/remotes/origin` only exists after checkout, so the `IsRemote` logic isn't relevant. The `IsBranch`/`IsTag` checks are handled on line 112 now:
```
if r.Name().String() == gs.ref || r.Name().Short() == gs.ref
```
That is, we take commits belonging to an exact-matching ref (e.g. `refs/heads/master`) or the first matching short ref (e.g. `master`).

EronWright (approved):

Issue #607, opened 2024-07-24
Implement graceful termination for the server, canceling outstanding operations


EronWright:
When the grpc server receives a shutdown signal (SIGINT), it should gracefully shutdown. It should not accept new RPC requests, and should wait for outstanding requests to complete.

Shutdown is usually due to a planned disruption, e.g. workspace pod termination due to a change to the project source or for node maintenance.

Ideally, any outstanding Pulumi deployment operation would be sent a CTRL-C signal to give it a chance to wind down gracefully. That's an unimplemented aspect of the core auto API: https://github.com/pulumi/pulumi/issues/13160

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

PR #659, opened 2024-09-11
[v2] Graceful shutdown

blampe:
This implements graceful shutdown by propagating interrupts to our child processes.

Killing a workspace pod in the middle of an update will mark the Update as failed ("update canceled" error), and subsequent Updates will resume where it left off as you would expect.

Fixes https://github.com/pulumi/pulumi-kubernetes-operator/issues/607.


agent/cmd/root.go
```diff
@@ -40,8 +40,8 @@ to use to perform stack operations.`,
 		zc := zap.NewDevelopmentConfig()
 		zc.DisableCaller = true
 		zc.DisableStacktrace = true
-		if !verbose {
-			zc.Level.SetLevel(zap.InfoLevel)
+		if verbose {
+			zc.Level.SetLevel(zap.DebugLevel)
```
blampe, line 44: The default level is already info, so debug seemed more appropriate for verbose.

agent/cmd/serve.go
```diff
@@ -105,38 +94,16 @@ var serveCmd = &cobra.Command{
...
 			log.Fatalw("unable to make an automation server", zap.Error(err))
 			os.Exit(1)
 		}
-
-		// Configure the grpc server.
-		// Apply zap logging and use filters to reduce log verbosity as needed.
-		address := fmt.Sprintf("%s:%d", Host, Port)
+		address := fmt.Sprintf("%s:%d", _host, _port)
 		log.Infow("starting the RPC server", "address", address)
-		serverOpts := []grpc_zap.Option{
```
EronWright, line 119: I felt better with this setup code being in serve.go rather than the server package, because a) the setup code tends to  be parameterized by flags, b) there might be other RPC services to be mixed into the server, and c) code that watches signals and calls os.Exit should be in the cmd package.  Moving the code has made this PR harder to understand.

EronWright (approved):
I do think the PR could and should be simplified to focus on graceful shutdown, but either way the improvement itself looks good.

agent/cmd/serve.go
```diff
@@ -16,54 +16,42 @@ limitations under the License.
...
-	WorkDir     string
-	SkipInstall bool
-	Stack       string
-	Host        string
-	Port        int
+	_workDir     string
+	_skipInstall bool
+	_stack       string
+	_host        string
+	_port        int
```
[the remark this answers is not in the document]
blampe: https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_

Makes global symbols (and their mutation) more obvious to readers.

PR #665, opened 2024-09-16
[v2] E2E test groundwork

blampe:
This adds the groundwork for e2e testing the v2 branch. Try it with `make test-e2e`.

* The Kubebuilder scaffolding was expecting some kustomize files on top of `operator/config`, so I regenerated CRDs into `operator/config/bases` and dropped a Deployment into `operator/config/manager`. We'll eventually need to consolidate this with the root `config/crd`.
* I happened to have a more recent version of `protoc-gen-go-grpc` which changed the generated code slightly. I checked in a `.mise.toml` mostly just to write down which versions of each tool I was using while working on this. Totally optional but you can read docs [here](http://mise.jdx.dev) if you'd like a direnv-style tooling setup.
* I removed the agent's Dockerfile, so we now have one image with manager and agent binaries.
* I added a make target to build the rootless repo. I think we'll still want to get this working with the root repo.
* I had to cut some corners with the image's version tag and how that gets plumbed to the agent. I'm currently generating a random hex tag and baking that into the binary to guarantee the operator spawns agents using the same image.
* I copy-pasted some of the scaffolded RBAC stuff to get leader election working. The operator doesn't do any cleanup so it should be safe to relinquish the lease on shutdown.
* Simplified the scaffolded e2e test -- we don't need prometheus or cert-manager (yet -- if we do we can add them later).

I haven't enabled an assertion for the random-yaml stack because I wasn't able to see it reliably reconcile. I was seeing GitHub rate limit errors as well as errors from the operator when updating status on the stack.


operator/config/crd/bases/auto.pulumi.com_updates.yaml
```diff
@@ -0,0 +1,217 @@
+---
```
blampe, line 1: Generated.

operator/config/crd/bases/pulumi.com_programs.yaml
```diff
@@ -0,0 +1,196 @@
+---
```
blampe, line 1: Generated.

operator/config/manager/manager.yaml
```diff
@@ -0,0 +1,65 @@
...
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: project
+    app.kubernetes.io/managed-by: kustomize
+  name: pulumi-kubernetes-operator
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
```
blampe, line 13: Kubebuilder scaffolding.

operator/config/rbac/leader_election_role.yaml
```diff
@@ -0,0 +1,40 @@
+# permissions to do leader election.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
blampe, line 3: This and all the other RBAC stuff is scaffolding.

operator/internal/controller/auto/workspace_controller.go
```diff
@@ -340,6 +338,9 @@ func labelsForStatefulSet(w *autov1alpha1.Workspace) map[string]string {
 }
 
 func newStatefulSet(ctx context.Context, w *autov1alpha1.Workspace, source *sourceSpec) (*appsv1.StatefulSet, error) {
+	// TODO: get from configuration
+	workspaceAgentImage := "pulumi/pulumi-kubernetes-operator-v2:" + version.Version
```
EronWright, line 342: Maybe this should be read from an environment variable.

EronWright (approved):

operator/e2e/e2e_test.go
```diff
@@ -0,0 +1,114 @@
...
+	if testing.Short() {
+		t.Skip()
+	}
+
+	// Generate a random tag to ensure pods get re-created when re-running the
+	// test locally.
+	bytes := make([]byte, 12)
+	_, _ = rand.Read(bytes) //nolint:staticcheck // Don't need crypto here.
+	tag := hex.EncodeToString(bytes)
+	projectimage := "pulumi/pulumi-kubernetes-operator-v2:" + tag
```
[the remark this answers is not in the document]
blampe: Using a `latest` tag with an `IfNotPresent` pull policy is sufficient for CI. I was having trouble running things multiple times locally because a new latest tag wouldn't cause things to re-deploy but I can probably simplify this with some more robust teardown logic.

operator/internal/controller/auto/workspace_controller.go
blampe: I'll try to hook this up to the downward API.

Issue #651, opened 2024-09-06
Implement the Program API


EronWright:
The Program API serves as way to write an inline YAML program to be deployed via the Stack API (see [docs](https://www.pulumi.com/docs/using-pulumi/continuous-delivery/pulumi-kubernetes-operator/#using-a-program-object)).

We need a way to feed the program to the workspace pod, during pod initialization. A couple of possible ways:
1. Have the init container get the `Program` object directly from the API server. This has an RBAC implication.
2. Have the program controller produce a downloadable artifact, akin to how `GitRepository` serves an artifact, and download the artifact as we do with Flux.  
 
I have a preference for (2). Note that the `Program` CRD doesn't have a status block as of yet, nor does it have a controller.

For reference, here's how the `GitRepository` CR works:
```yaml
apiVersion: source.toolkit.fluxcd.io/v1
kind: GitRepository
metadata:
  creationTimestamp: "2024-07-25T17:06:15Z"
  finalizers:
    - finalizers.fluxcd.io
  generation: 1
  name: pulumi-examples
  namespace: default
  resourceVersion: "2005488"
  uid: 1b67a048-8ec1-4486-8fcc-79858a48fa35
spec:
  interval: 5m0s
  ref:
    branch: master
  timeout: 60s
  url: https://github.com/pulumi/examples.git
status:
  artifact:
    digest: sha256:bcbed45526b241ab3366707b5a58c900e9d60a1d5c385cdfe976b1306584b454
    lastUpdateTime: "2024-09-04T17:32:18Z"
    path: gitrepository/default/pulumi-examples/f143bd369afcb5455edb54c2b90ad7aaac719339.tar.gz
    revision: master@sha1:f143bd369afcb5455edb54c2b90ad7aaac719339
    size: 48988266
    url: http://source-controller.flux-system.svc.cluster.local./gitrepository/default/pulumi-examples/f143bd369afcb5455edb54c2b90ad7aaac719339.tar.gz
  conditions:
    - lastTransitionTime: "2024-09-05T21:12:31Z"
      message: stored artifact for revision 'master@sha1:f143bd369afcb5455edb54c2b90ad7aaac719339'
      observedGeneration: 1
      reason: Succeeded
      status: "True"
      type: Ready
    - lastTransitionTime: "2024-09-04T17:32:18Z"
      message: stored artifact for revision 'master@sha1:f143bd369afcb5455edb54c2b90ad7aaac719339'
      observedGeneration: 1
[...]

PR #673, opened 2024-09-20
[V2] Program controller and file server for program objects

rquitales:
### Proposed changes

This PR enables a Workspace pod to fetch a Program object in a similar fashion to how it fetches Flux artifacts. We do this by exposing a HTTP server that serves tarballs of a fully formed `Pulumi.yaml` file, incorporating the requested Program spec. Unlike the approach with the Flux source-controller, I've opted **not to** create/store these tar files in local ephemeral storage in the controller pod. Instead, when the artifact URL is accessed, the file server will fetch the requested Program resource from the Kubernetes API server, and wrap it up in a `Pulumi.yaml` file and tarred. This approach ensures that the most recent Program spec is always served, and it also greatly simplifies storage since we do not need to create a local duplicate of these Program objects. Since the source of truth is always on cluster, we do not need to continuously reconcile and generate new artifacts for new generations of Programs. Should we choose to change this implementation in the future, the current strategy using the status field to convey the URL should make it easy to do so.

- [x] Add status field to the Program resource to advertise a downloadable URL for the program
- [x] Scaffold a program-controller to reconcile the status/URL
- [x] Create a simple file server to serve the fully-formed Pulumi.yaml from a Program URL
- [x] Update deployment manifests to expose the file server
- [x] Additional unit tests
- [x] Rebase PR to take in test changes
- [x] Integrate with stack-controller

### Related issues (optional)

Closes: #651


operator/api/pulumi/v1/program_types.go
```diff
@@ -160,16 +160,35 @@ type Getter struct {
...
+	// Artifact represents the last succesful artifact generated by program reconciliation.
+	// +optional
+	Artifact *Artifact `json:"artifact,omitempty"`
+}
+
 // +kubebuilder:object:root=true
 // +kubebuilder:storageversion
+// +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].message",description=""
```
EronWright, line 183: The Program doesn't have a Ready condition, assumedly you had planned to before going "on-demand".  Maybe it still should have a ready condition, if you would want the stack controller to wait for it.  

I suppose the Program's generation field is sufficient for detecting a program change.

operator/cmd/main.go
```diff
@@ -127,6 +143,16 @@ func main() {
...
 
+	// Create a new ProgramHandler to handle Program objects.
+	pHandler := newProgramHandler(mgr.GetClient(), programFSAdvAddr)
+	// Start a simple file-server to serve CR files as compressed tarballs.
+	go func() {
+		// Wait until our manager is elected leader before starting the file server.
+		<-mgr.Elected()
+
+		startProgramFileServer(pHandler, programFSAddr)
+	}()
```
EronWright, line 154: nit: I think you can add arbitrary runnables to the manager via its `Add` method, and it will start/stop the runnable like it does the controllers...assumedly after leader election.

operator/cmd/main.go
```diff
@@ -168,3 +203,55 @@ func main() {
...
+
+// startProgramFileServer starts a simple file server to serve Program objects as compressed tarballs. This allows
+// children pods with restricted permissions to access the Program objects, without needing read permissions granted.
+func startProgramFileServer(programHandler *pulumi.ProgramHandler, address string) {
+	setupLog.Info("starting file server to serve Program objects", "address", address, "advertisedAddress", programHandler.Address())
+	mux := http.NewServeMux()
+	mux.Handle("/programs/", programHandler.HandleProgramServing())
+	err := http.ListenAndServe(address, mux)
+	if err != nil {
+		setupLog.Error(err, "Program file server error")
```
EronWright, line 215: The error should be fatal.

operator/cmd/main.go
```diff
@@ -168,3 +203,55 @@ func main() {
...
 	}
 }
+
+// startProgramFileServer starts a simple file server to serve Program objects as compressed tarballs. This allows
+// children pods with restricted permissions to access the Program objects, without needing read permissions granted.
+func startProgramFileServer(programHandler *pulumi.ProgramHandler, address string) {
+	setupLog.Info("starting file server to serve Program objects", "address", address, "advertisedAddress", programHandler.Address())
+	mux := http.NewServeMux()
+	mux.Handle("/programs/", programHandler.HandleProgramServing())
+	err := http.ListenAndServe(address, mux)
```
EronWright, line 213: When the operator is terminated, can we call Shutdown to allow an outstanding request to complete?

operator/cmd/main.go
```diff
@@ -39,6 +38,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
+	autov1alpha1 "github.com/pulumi/pulumi-kubernetes-operator/operator/api/auto/v1alpha1"
+	pulumiv1 "github.com/pulumi/pulumi-kubernetes-operator/operator/api/pulumi/v1"
+	autocontroller "github.com/pulumi/pulumi-kubernetes-operator/operator/internal/controller/auto"
+	"github.com/pulumi/pulumi-kubernetes-operator/operator/internal/controller/pulumi"
```
EronWright, line 45: Please use the alias `pulumicontroller` for this package.

operator/config/manager/kustomization.yaml
```diff
@@ -3,6 +3,5 @@ kind: Kustomization
 images:
 - name: controller
   newName: pulumi/pulumi-kubernetes-operator-v2
-  newTag: latest
```
EronWright, line 6: Is this intentional?

operator/internal/controller/pulumi/program_controller.go
```diff
@@ -0,0 +1,251 @@
...
+//+kubebuilder:rbac:groups=pulumi.com,resources=programs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=pulumi.com,resources=programs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=pulumi.com,resources=programs/finalizers,verbs=update
+
+// Reconcile reconciles the Program object.
+func (r *ProgramReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := log.FromContext(ctx)
+	log.Info("Reconciling Program")
+
+	// Feth the Program object.
```
EronWright, line 209:
```suggestion
	// Fetch the Program object.
```

operator/internal/controller/pulumi/program_controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+			Expect(program.Status.Artifact.URL).
+				To(Equal(fmt.Sprintf("%s/programs/%s/%s", advertisedAddress, programNamespacedName.Namespace, programNamespacedName.Name)))
+
+			By("reconciling the same Program object but with a new spec change")
+			program.Program.Resources = map[string]v1.Resource{
+				"test-resource": {
+					Type: "kubernetes:core/v1:Service",
+				},
+			}
+			program.ObjectMeta.Generation++
```
EronWright, line 112: Why are you incrementing the generation manually? The Update call itself will do that.

EronWright (changes requested):

operator/config/manager/kustomization.yaml
rquitales: I don't think so. Have reverted this change.

operator/api/pulumi/v1/program_types.go
rquitales: Since we're doing this 'on-demand', having a ready condition doesn't add much value. To keep the status block simple for now, we can use the `.status.artifact.url` field to detect when the program can be downloaded.

operator/api/pulumi/v1/program_types.go
```diff
@@ -160,16 +160,35 @@ type Getter struct {
...
+// StackStatus defines the observed state of Program.
+type ProgramStatus struct {
+	// ObservedGeneration is the last observed generation of the Program
+	// object.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// LastResyncTime contains a timestamp for the last time a resync of the stack took place.
+	// +optional
+	LastResyncTime metav1.Time `json:"lastResyncTime,omitempty"`
```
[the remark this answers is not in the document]
EronWright: I still think this field should be removed because a) it is redundant given the observedGeneration, and b) it uses the term "sync" which has a particular (different) meaning in PKO.

operator/internal/controller/pulumi/program_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	program := new(pulumiv1.Program)
+	if err := r.Get(ctx, req.NamespacedName, program); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Update the status of the Program object to contain an updated URL.
+	program.Status.Artifact = &pulumiv1.Artifact{
+		URL: r.ProgramHandler.CreateProgramURL(program.Namespace, program.Name),
+	}
+	program.Status.LastResyncTime = metav1.Now()
```
EronWright, line 258: This line will make the object be more volatile than necessary.

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -34,6 +34,7 @@ import (
 	autov1alpha1 "github.com/pulumi/pulumi-kubernetes-operator/operator/api/auto/v1alpha1"
 	"github.com/pulumi/pulumi-kubernetes-operator/operator/api/pulumi/shared"
 	pulumiv1 "github.com/pulumi/pulumi-kubernetes-operator/operator/api/pulumi/v1"
+	v1 "github.com/pulumi/pulumi-kubernetes-operator/operator/api/pulumi/v1"
```
EronWright, line 37: Looks to be redundant.

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -628,24 +629,35 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
+		}
+
+		// The Program.status is setup to mimic a FluxSource, so we can use the same function to
+		// initiate the workspace.
+		currentCommit, err = sess.SetupWorkspaceFromFluxSource(ctx, program, &shared.FluxSource{})
+		if err != nil {
+			if isStalledError(err) {
+				instance.Status.MarkStalledCondition(pulumiv1.StalledCrossNamespaceRefForbiddenReason, err.Error())
+				return reconcile.Result{}, saveStatus()
+			}
```
EronWright, line 657: You can remove this special case, because we decoupled the generic workspace setup code (it is now executed in "step 2").

EronWright (approved):
Looks great, just left a couple of comments.

operator/internal/controller/pulumi/program_controller.go
```diff
@@ -0,0 +1,314 @@
...
+type ProjectFile struct {
+	Name    string `json:"name"`
+	Runtime string `json:"runtime"`
+	pulumiv1.ProgramSpec
+
+	timeModified metav1.Time // this field is not to be serialized
+}
+
+// tarProgram creates a tarball of the Pulumi project file and writes it to the provided writer.
+func tarProgram(data []byte) ([]byte, error) {
```
blampe, line 66: Nit: you should be able to chain the gzip writer instead of using a second buffer. I don't _think_ the info header needs to know the size upfront. Something like this

```go
gzw := gzip.NewWriter(buf)
defer gzw.Close()
tw := tar.NewWriter(gzw)
defer tw.Close()
```

But these files are so small it doesn't really matter.

Edit: and if you keep it all in terms of readers/writers you can use the `ResponseWriter` directly so no need for a buffer.

operator/internal/controller/pulumi/program_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	buf := new(bytes.Buffer)
```
blampe, line 67:
```suggestion
	buf := bytes.Buffer{}
```

operator/internal/controller/pulumi/program_controller.go
```diff
@@ -0,0 +1,314 @@
...
+
+	tarData, err := tarProgram(data)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to create tarball of project file: %w", err)
+	}
+	gzData, err := gzipFile(tarData)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to compress tarball: %w", err)
+	}
+	program.Status.Artifact.Digest = calculateDigest(gzData)
```
blampe, line 278: Maybe a better reason to keep tar/gz in terms of readers/writers -- Hash implements Writer so you can use it in place of a ResponseWriter.

operator/internal/controller/pulumi/program_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	stackInformer, err := mgr.GetCache().GetInformer(context.Background(), new(pulumiv1.Program))
+	if err != nil {
+		return err
+	}
+	if _, err = stackInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    newProgramCallback,
+		DeleteFunc: deleteProgramCallback,
+	}); err != nil {
+		return err
+	}
```
blampe, line 306: Nice!

operator/internal/controller/pulumi/program_controller.go
```diff
@@ -0,0 +1,314 @@
...
+
+	// Create a new controller and set the options. We will only reconcile on spec changes to the Program resource (new generation).
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(ProgramControllerName).
+		For(new(pulumiv1.Program)).
```
blampe, line 311: Nit
```suggestion
		For(pulumiv1.Program{}).
```

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -628,24 +629,35 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
-		// 			return reconcile.Result{}, nil
-		// 		}
-		// 		instance.Status.MarkReconcilingCondition(pulumiv1.ReconcilingRetryReason, err.Error())
-		// 		// this can fail for reasons which might go away without intervention; so, retry explicitly
-		// 		return reconcile.Result{Requeue: true}, nil
-		// 	}
+	case stack.ProgramRef != nil:
+		var program unstructured.Unstructured
+		program.SetAPIVersion(v1.GroupVersion.String())
+		program.SetKind("Program")
```
blampe, line 635: I'm curious why Unstructured and not the Program type here?

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -628,24 +629,35 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
+				instance.Status.MarkStalledCondition(pulumiv1.StalledSourceUnavailableReason, errProgramNotFound.Error())
+				return reconcile.Result{}, saveStatus()
+			}
+			log.Error(err, "Failed to get Program object", "Name", stack.ProgramRef.Name)
+			return reconcile.Result{}, err
+		}
+
+		// The Program.status is setup to mimic a FluxSource, so we can use the same function to
+		// initiate the workspace.
+		currentCommit, err = sess.SetupWorkspaceFromFluxSource(ctx, program, &shared.FluxSource{})
```
blampe, line 652: Interesting, what's the currentCommit in this case - the digest?

blampe (approved):
Looks good, mostly just some nits but nothing blocking.

operator/api/pulumi/v1/program_types.go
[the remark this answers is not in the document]
rquitales: - `field should be removed`: I'm opposed to this more so because we can have multiple reconciliation on the same generation. Take the example where a Program object has been reconciled and the status is written to it. If the controller pod restarts, there might be a possibility that the hostname has changed (though very unlikely). When the controller pod restarts, it will reconcile all objects it should watch for, and thus, the URL stored in status will be updated. It would be helpful for debugging purposes to know when the URL was generated.
- `the term "sync" which has a particular (different) meaning in PKO.`: yes, a different name for this field, should we ultimately decide to keep it, should be used!

operator/internal/controller/pulumi/stack_controller.go
rquitales: The `SetupWorkspaceFromFluxSource` method takes in an unstructured object to obtain the `URL` and `Digest` fields. Instead of getting the object as a Program and then converting to unstructured (ie. `runtime.DefaultUnstructuredConverter.ToUnstructured(object)`), I'm just doing this directly.

operator/internal/controller/pulumi/stack_controller.go
rquitales: Yes, the digest/hashed tar file.

operator/internal/controller/pulumi/program_controller.go
rquitales: Updated to just combine the 2 functions by chaining the gzip/tar writers and use read/writers instead of a buffer.

operator/api/pulumi/v1/program_types.go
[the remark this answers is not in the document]
rquitales: I've removed this field for now, to merge this PR in eagerly. We can rethink how we want to handle this and add this in the future if we choose to.

operator/internal/controller/pulumi/program_controller.go
rquitales: Removed, along with the `lastResyncTime` field.

PR #678, opened 2024-09-20
[v2] Stable envRefs

blampe:
Stack CRDs that include `envRefs` will currently thrash the workspace's underlying StatefulSet because map order is undefined.

This sorts the envRef keys ahead of time to keep the StatefulSet stable. As a result we're also able to turn on the E2E test in CI since it should now succeed reliably.

### DIFF
--- a/operator/e2e/e2e_test.go
+++ b/operator/e2e/e2e_test.go
@@ -91,9 +91,6 @@ func TestE2E(t *testing.T) {
 
 				cmd := exec.Command("bash", "-c", "envsubst < e2e/testdata/git-auth-nonroot/* | kubectl apply -f -")
 				require.NoError(t, run(cmd))
-				t.Cleanup(func() {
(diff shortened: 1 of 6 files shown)


operator/e2e/e2e_test.go
```diff
@@ -91,9 +91,6 @@ func TestE2E(t *testing.T) {
 
 				cmd := exec.Command("bash", "-c", "envsubst < e2e/testdata/git-auth-nonroot/* | kubectl apply -f -")
 				require.NoError(t, run(cmd))
-				t.Cleanup(func() {
-					_ = run(exec.Command("kubectl", "delete", "-f", "e2e/testdata/git-auth-nonroot"))
-				})
```
blampe, line 96: This was deleting the secret needed by the operator in order to poll the repo, preventing the stack from reconciling / getting deleted.

EronWright (approved):

PR #679, opened 2024-09-21
[V2] Fix log spamming from cache misses when updating status repeatedly

rquitales:
### Proposed Changes

This PR addresses the issue of excessive log spam caused by reconcile loops being triggered in quick succession on the same Workspace object due to changes in its dependent resources (e.g., services, statefulsets). Previously, each reconcile loop attempted to update the Status subresource, but because the fetched object had an outdated `resourceVersion` (due to a stale cache), it would result in a conflict error. 

While `r.Update()` invalidates the cache, `r.Status().Update()` does not, leading to this conflict when only the status was updated in consecutive reconcile loops. The failed reconcile loop would eventually succeed once the cache was updated, but the log spam remained an issue.

This PR resolves the problem by patching the status subresource instead of performing an update, which avoids the need for the latest `resourceVersion` and prevents the conflict errors from occurring.

Refs #700

### Commits
1. Patch status subresource instead of using Update to prevent conflicts caused by cache miss.
2. Add a field manager to all Kubernetes API calls.

### DIFF
--- a/operator/internal/controller/auto/workspace_controller.go
+++ b/operator/internal/controller/auto/workspace_controller.go
@@ -85,6 +85,8 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 		return ctrl.Result{}, err
 	}
+	// Store a copy of the original workspace to create patches for status updates.
+	originalWorkspace := w.DeepCopy()
 
 	ready := meta.FindStatusCondition(w.Status.Conditions, WorkspaceConditionTypeReady)
 	if ready == nil {
@@ -97,7 +99,9 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		w.Status.ObservedGeneration = w.Generation
 		ready.ObservedGeneration = w.Generation
 		meta.SetStatusCondition(&w.Status.Conditions, *ready)
-		return r.Status().Update(ctx, w)
+		// Use .Status().Patch() instead of .Update() to avoid conflicts with sequential reconcile loops hitting a
+		// stale version of the object from the cache.
+		return r.Status().Patch(ctx, w, client.MergeFrom(originalWorkspace), client.FieldOwner(FieldManager))
 	}
 
 	if w.DeletionTimestamp != nil {
--- a/operator/internal/controller/auto/update_controller.go
+++ b/operator/internal/controller/auto/update_controller.go
@@ -106,7 +106,7 @@ func (r *UpdateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		meta.SetStatusCondition(&obj.Status.Conditions, *rs.failed)
 		rs.complete.ObservedGeneration = obj.Generation
 		meta.SetStatusCondition(&obj.Status.Conditions, *rs.complete)
-		return r.Status().Update(ctx, obj)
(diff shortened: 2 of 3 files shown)


operator/internal/controller/auto/workspace_controller.go
```diff
@@ -97,7 +97,9 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		w.Status.ObservedGeneration = w.Generation
 		ready.ObservedGeneration = w.Generation
 		meta.SetStatusCondition(&w.Status.Conditions, *ready)
-		return r.Status().Update(ctx, w)
+		// Use .Status().Patch() instead of .Update() to avoid conflicts with sequential reconcile loops hitting a
+		// stale version of the object from the cache.
+		return r.Status().Patch(ctx, w, client.MergeFrom(w), client.FieldOwner(FieldManager))
```
rquitales, line 102: We could alternatively do something like the following if we wanted to keep using `.Update` and reduce log spam:
```
err := r.Status().Update(...)
if err != nil && errors.IsConflict {
    return ctrl.Result{Requeue: true}, nil
}
return ctrl.Result{}, err
```

operator/internal/controller/auto/workspace_controller.go
```diff
@@ -97,7 +99,9 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		w.Status.ObservedGeneration = w.Generation
 		ready.ObservedGeneration = w.Generation
 		meta.SetStatusCondition(&w.Status.Conditions, *ready)
-		return r.Status().Update(ctx, w)
+		// Use .Status().Patch() instead of .Update() to avoid conflicts with sequential reconcile loops hitting a
+		// stale version of the object from the cache.
+		return r.Status().Patch(ctx, w, client.MergeFrom(originalWorkspace), client.FieldOwner(FieldManager))
```
rquitales, line 104: We could alternatively do something like the following if we wanted to keep using `.Update` and reduce log spam:
```
err := r.Status().Update(...)
if err != nil && errors.IsConflict {
    return ctrl.Result{Requeue: true}, nil
}
return ctrl.Result{}, err
```

blampe: Looking at [UpdateStatus](https://github.com/kubernetes/client-go/blob/def003bc9552f7e98726f5f2d05a701d33300418/gentype/type.go#L252-L265) makes me think patching (or more specifically put'ing) the status is fairly typical, so I'm OK with this.

blampe (approved):

EronWright (changes requested):
I think this could lead to `.status.observedGeneration` regressing, e.g. from 2 to 1, if the object is cached.

My theory of the root cause is:
1. The reconciler is seeing a conflict when attempting to write the status, due to optimistic concurrency.
2. The reconciler may see a cached object if reconciliation runs just after a write. There's no guarantee that, when reconciliation begins, that the cache of stacks is fresh.
3. Writes to the API Server do not update the cache.
4. There's many watch triggers for other objects, that cause a stack to be enqueued for reconciliation. 

i would prefer we try to tune the watches on the child objects, to be less aggressive in enqueuing the owner. For example, we watch the StatefulSet but are we filtering on the relevant status change? Basically the system is racing between the status change on the owner and any change on the owned.

PR #686, opened 2024-09-25
[v2] Consolidate go.mod

blampe:
* Consolidate `{operator,agent}/go.mod` under a root `go.mod`.
* Rewrite imports to use v2 path.
* Remove `/test` -- these were still referring to v1 code and can be revived in a followup if we want to keep any of them.
* Move Dockerfile to repo root and fix image build.

Makefile targets are unchanged except for a small rename docker-rootless → docker-nonroot.

Fixes https://github.com/pulumi/pulumi-kubernetes-operator/issues/687.


Dockerfile
```diff
@@ -25,38 +21,36 @@ COPY . /operator/.
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
 
 ARG VERSION
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager -ldflags="-X github.com/pulumi/pulumi-kubernetes-operator/operator/version.Version=${VERSION}" cmd/main.go
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o /manager -ldflags="-X github.com/pulumi/pulumi-kubernetes-operator/v2/operator/version.Version=${VERSION}" ./operator/cmd/main.go
 
 # Build the agent binary
-FROM golang:1.22 AS agent-builder
+FROM golang:1.23 AS agent-builder
```
EronWright, line 27:
```suggestion
FROM --platform=${BUILDPLATFORM} golang:1.23 AS agent-builder
```

EronWright (approved):
I agree with removing the `tests/` folder because it seems obsolete.

EronWright (changes requested):
I just saw that the Dockerfile was moved, which is fine but please update the Makefile and goreleaser file accordingly. 

There's targets `docker-build` `docker-buildx` and `bundle` to consider.  I suppose you could move some stuff down to the root makefile.  The 'bundle' images are, I suppose, most obviously tied to the operator and `bundle` seem to belong in operator makefile.

blampe:
> I just saw that the Dockerfile was moved, which is fine but please update the Makefile and goreleaser file accordingly.
> 
> There's targets `docker-build` `docker-buildx` and `bundle` to consider. I suppose you could move some stuff down to the root makefile. The 'bundle' images are, I suppose, most obviously tied to the operator and `bundle` target seems to belong in operator makefile.

Good call @EronWright I fixed pointed goreleader and `docker-buildx` at the new location. I think it would make sense to eventually move most (all?) of the operator's Makefile to the root but it's not urgent.

operator/Makefile
```diff
@@ -171,10 +171,7 @@ docker-push: ## Push docker image with the manager.
 PLATFORMS ?= linux/arm64,linux/amd64
 .PHONY: docker-buildx
 docker-buildx: ## Build and push docker image for the manager for cross-platform support
-	- $(CONTAINER_TOOL) buildx create --name project-v3-builder
-	$(CONTAINER_TOOL) buildx use project-v3-builder
-	- $(CONTAINER_TOOL) buildx build --push --platform=$(PLATFORMS) --build-context agent=../agent --tag ${IMG}:${VERSION} -f Dockerfile .
-	- $(CONTAINER_TOOL) buildx rm project-v3-builder
+	- $(CONTAINER_TOOL) buildx build --push --platform=$(PLATFORMS) --tag ${IMG}:${VERSION} ..
```
EronWright, line 174:
```suggestion
	- $(CONTAINER_TOOL) buildx build --push --platform=$(PLATFORMS) --tag ${IMG}:${VERSION} --build-arg VERSION=${VERSION} ..
```

.goreleaser.yml
```diff
@@ -74,7 +74,7 @@ dockers:
     goarch: amd64
 
     # Path to the Dockerfile (from the project root).
-    dockerfile: operator/Dockerfile
+    dockerfile: Dockerfile
```
EronWright, line 77: Note that goreleaser doesn't provide the necessary context or VERSION parameter for our Dockerfile, so this is probably not working.

EronWright (approved):

.goreleaser.yml
blampe: Yeah I also realized we're skipping the docker step during the dry run?

EronWright: Maybe we should remove the `dockers` section of the goreleaser if it isn't expected to work.

Issue #687, opened 2024-09-25
Align the go.mod files


EronWright:
Combine the go mods into a singular mod file at the root, using go 1.23, and using a mod name of `github.com/pulumi/pulumi-kubernetes-operator/v2`.

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

PR #688, opened 2024-09-25
fixes to push docker image

EronWright:

### DIFF
--- a/operator/Dockerfile
+++ b/operator/Dockerfile
@@ -1,5 +1,5 @@
 # Build the operator binary
-FROM --platform=${BUILDPLATFORM} golang:1.23 AS op-builder
+FROM golang:1.23 AS op-builder
 ARG TARGETOS
 ARG TARGETARCH
 
--- a/operator/Dockerfile
+++ b/operator/Dockerfile
@@ -1,5 +1,5 @@
 # Build the operator binary
-FROM golang:1.23 AS op-builder
+FROM --platform=${BUILDPLATFORM} golang:1.23 AS op-builder
 ARG TARGETOS
 ARG TARGETARCH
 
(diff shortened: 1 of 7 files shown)


operator/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the operator binary
-FROM --platform=${BUILDPLATFORM} golang:1.23 AS op-builder
+FROM golang:1.23 AS op-builder
```
blampe: You want to keep this -- it lets you build multi-platform images natively. If you're having platform issues it might be because we're missing this on the other build targets.

operator/Dockerfile
```diff
@@ -28,7 +28,7 @@ ARG VERSION
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager -ldflags="-X github.com/pulumi/pulumi-kubernetes-operator/operator/version.Version=${VERSION}" cmd/main.go
 
 # Build the agent binary
-FROM golang:1.22 AS agent-builder
+FROM --platform=${BUILDPLATFORM} golang:1.22 AS agent-builder
```
blampe, line 31:
```suggestion
FROM --platform=${BUILDPLATFORM} golang:1.23 AS agent-builder
```

operator/e2e/e2e_test.go
```diff
@@ -45,7 +45,7 @@ func TestE2E(t *testing.T) {
 	b := make([]byte, 12)
 	_, _ = rand.Read(b) //nolint:staticcheck // Don't need crypto here.
 	tag := hex.EncodeToString(b)
-	projectimage := "pulumi/pulumi-kubernetes-operator-v2:" + tag
+	projectimage := "pulumi/pulumi-kubernetes-operator:" + tag
```
blampe, line 48: I worry about folks currently using `pulumi/pulumi-kubernetes-operator:latest`. This has the potential to break people, or at least require some manual migration steps -- assuming we eventually publish v2 to latest.

operator/Dockerfile
EronWright: For whatever reason the `make docker-buildx` target was in the habit of adding `--platform=${BUILDPLATFORM}` automatically, but I see no reason to do that and changed it.

operator/Dockerfile
EronWright: I thought about changing it, but the agent's go.mod is indeed 1.22 so I figured we'd defer to https://github.com/pulumi/pulumi-kubernetes-operator/issues/687

operator/e2e/e2e_test.go
EronWright: I sympathize. My plan is to NOT apply a suffix, but also NOT push to latest.

Issue #690, opened 2024-09-26
Add single-namespace deployment mode


EronWright:

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

EronWright:
Closes: https://github.com/pulumi/pulumi-kubernetes-operator/issues/279

PR #691, opened 2024-09-26
Simplified deployment app

EronWright:
### Proposed changes

This is a new, simplified deployment app for PKO.  It leverages the kustomization that is maintained in `operator/config/default` to avoid code duplication.  It supports ONLY cluster-wide installation.

PKOv2 is designed to natively support multi-tenancy across namespaces without needing to deploy separate instances of the operator into each namespace. Looking back at https://github.com/pulumi/pulumi-kubernetes-operator/pull/328, we see that the ability to deploy to multiple namespaces was added to address a lack of isolation (see https://github.com/pulumi/home/issues/2330). I opened https://github.com/pulumi/pulumi-kubernetes-operator/issues/690 to track adding support for single-namespace deployment.


operator/Dockerfile
```diff
@@ -49,6 +49,7 @@ COPY --from=agent / .
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
+ARG VERSION
```
EronWright, line 52: A bugfix since ARG is per-stage.

deploy/deploy-operator-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,13 @@
...
+  Deploys the Pulumi Kubernetes Operator.
+config:
+  version: # The version to install of the Pulumi Kubernetes Operator.
+    type: string
+    default: v2.0-devel
+resources:
+  pko:
+    type: kubernetes:kustomize/v2:Directory
+    properties:
+      directory: https://github.com/pulumi/pulumi-kubernetes-operator//operator/config/default/?ref=${version}
```
rquitales, line 13: nit: remove extra `/`
```suggestion
      directory: https://github.com/pulumi/pulumi-kubernetes-operator/operator/config/default/?ref=${version}
```

rquitales (approved):

deploy/deploy-operator-yaml/Pulumi.yaml
EronWright: The double-slash is by design: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md#remote-directories

>  The directory is specified by appending a // after the repo URL.

blampe (approved):

Issue #653, opened 2024-09-07
Add non-root user account to Pulumi image


EronWright:
For the pu/pu image to support non-root execution (a security best practice, and supported by PKOv2), we need to add a local user to the image.

Ideally we wouldn't have separate "nonroot" images (see [the "distroless" images](https://github.com/GoogleContainerTools/distroless?tab=readme-ov-file#what-images-are-available)).  Otherwise the defaulting logic would be more tricky.

One possibility is to add the local user as shown below but avoid using the `USER` instruction. I think that's sufficient for purposes of PKOv2, because the pod can apply a security context. The `USER` instruction seems to set the default, which might be considered a breaking change for existing users of the pu/pu image.

Here's an example of the Dockerfile commands:
```Dockerfile
ARG BASE_IMAGE=pulumi/pulumi:latest
FROM ${BASE_IMAGE}
ARG UID=1000
ARG GID=1000
RUN addgroup --gid $GID pulumi && \
    adduser --uid $UID --gid $GID --disabled-password --gecos "" pulumi && \
    echo 'pulumi ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
USER pulumi:pulumi
```

And an example of the pod security context:
```yaml
  securityContext:
    runAsGroup: 1000
    runAsNonRoot: true
    runAsUser: 1000
```

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

EronWright:
TODO: add `securityProfile` to stack spec, to allow the workspace's profile to be configured.

EronWright:
A nonroot variant of the kitchen sink image will be available soon:
https://github.com/pulumi/pulumi-docker-containers/pull/277

I believe the remaining work on the PKO side is to use the correct default based on the security profile:
- "restricted" - `pulumi/pulumi:latest-nonroot`
- "baseline" - `pulumi/pulumi:latest`

julienp:
The `latest-nonroot` tag is now published https://hub.docker.com/r/pulumi/pulumi/tags

rquitales:
Resolved by #693, #692 and https://github.com/pulumi/pulumi-docker-containers/pull/277.

PR #693, opened 2024-09-26
[V2] Set default statefulset image based on securityProfile

rquitales:
### Proposed changes

This update removes the default CRD validation marker from the `workspace.spec.image` field, allowing for dynamic image selection based on the value of `workspace.spec.securityProfile`. When `securityProfile` is set to `baseline`, the image defaults to `pulumi/pulumi:latest`. If `securityProfile` is marked as `restricted`, the system will instead select `pulumi/pulumi:latest-nonroot`. If a user specifies their own image, then that value is used instead. Handling these default settings within the controller is necessary, as CRD validation markers cannot accommodate conditional defaults.

Before implementing this feature, additional ginkgo tests were introduced to confirm that the desired behavior is properly achieved.

### Related issues (optional)

Closes: #653


EronWright (approved):
fabulous

blampe (approved):

Issue #576, opened 2024-05-02
Add metrics for currently reconciling stacks


nicu-da:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

### Affected area/feature
Operator metrics

Add a metric that is similar to the stacks_failing 

```
stacks_failing - a set of gauge time series, labelled by namespace, that gives the number of stacks currently failing (stack.status.lastUpdate.state is failed)
```

that tracks the stacks that are currently being reconciled. 

Compared to `controller_runtime_active_workers` the new metric should be labeled with the stack name, thus allowing visualization for what metrics are currently being updated, and to add alerts if a stack takes too long to update.

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

PR #706, opened 2024-10-04
[V2] Enable metrics server

rquitales:
### Proposed changes

This PR exposes the metrics service in our manifests and adds new unit tests to ensure the metrics logic is correct.

### Technical changes

1. Added `sigs.k8s.io/controller-runtime/pkg/metrics/filters` dependency to allow metrics to be secure by default (guidance from kubebuilder book/scaffold)
2. Adds new kustomize patches and bases to expose the metrics server in a typical PKO deployment
3. Refactor the `updateStackCallback` function
4. Added the `stacks_reconciling` metric
5. Added ginkgo tests to exercise the Program and Stack metrics
6. Fixed some bugs in our metrics as determined through the unit tests

### Testing

- Manually tested a deployed PKO installation with the Prometheus Operator
- Added unit tests (using ginkgo)

Note, e2e tests will be added in a followup PR.

### Related issues (optional)

Closes: #576


EronWright (approved):
Really cool.

The "filter" seems to be a perfect compliment to our plan for authenticating the Automation RPC endpoint. It imports the "apiserver" library and that will probably be useful for that effort.

go.mod
```diff
@@ -29,6 +29,29 @@ require (
 	sigs.k8s.io/controller-runtime v0.18.4
 )
 
+require (
```
blampe, line 32: Nit -

There should only be two `require` blocks (for direct and indirect) but I'm guessing go is getting confused by the replace directive? I wonder if `go mod tidy` consolidates these after the `replace` is moved to the top or bottom of the file.

operator/cmd/main.go
```diff
@@ -75,13 +76,14 @@ func main() {
...
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metric endpoint binds to. "+
+		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.BoolVar(&secureMetrics, "metrics-secure", false,
-		"If set the metrics endpoint is served securely")
+	flag.BoolVar(&secureMetrics, "metrics-secure", true,
```
blampe, line 85: 👍

operator/config/manager/manager.yaml
```diff
@@ -43,7 +43,6 @@ spec:
           args:
             - --leader-elect
             - --health-probe-bind-address=:8081
-            - --metrics-bind-address=:8383
```
blampe, line 46: I think there's an argument for keeping the port as-is considering the v1 chart spits out a ServiceMonitor that targets [:8383](https://github.com/pulumi/pulumi-kubernetes-operator/blob/d02ae39ed293efd2d779f3a10a530944ea054338/deploy/helm/pulumi-operator/templates/servicemonitor.yaml#L16) instead of the port name. It would be nice for the user's metrics to Just Work™ -- although if we're changing up the keys significantly then maybe there's no point.

In any case we should probably try to keep the metrics service as close as possible to what we have in v1, to at least keep selectors the same.

Maybe unrelated - it seems like the v1 operator has _two_ metrics ports, 8383 and 8686? Do you know why that is?

operator/internal/controller/pulumi/metrics.go
```diff
@@ -1,77 +0,0 @@
...
-			Help: "Number of stacks currently registered where the last reconcile failed",
-		},
-		[]string{"namespace", "name"},
-	)
-
-	collectors = append(collectors, numStacks, numStacksFailing)
-	return collectors
-}
-
-func init() {
```
blampe, line 36: 🙌 🙌 🙌

operator/internal/controller/pulumi/metrics_program_test.go
```diff
@@ -0,0 +1,52 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pulumi
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
```
blampe, line 19: I still don't like the linting sacrifice we take on by using these libraries. Especially in this case I think it makes things harder to read and more complicated than necessary. You're free to take this or leave it!

```go
package pulumi

import (
	"testing"

	"github.com/prometheus/client_golang/prometheus/testutil"
	"github.com/stretchr/testify/assert"
)

func TestProgramMetrics(t *testing.T) {
	numPrograms.Set(0)

	// it should increment the programs_active metric when a new Program is created
	newProgramCallback(nil)
	assert.Equal(t, 1.0, testutil.ToFloat64(numPrograms))

	// it should increment the programs_active metric when another Program is created
	newProgramCallback(nil)
	assert.Equal(t, 2.0, testutil.ToFloat64(numPrograms))

	// it should decrement the programs_active metric when a Program is deleted
	deleteProgramCallback(nil)
	assert.Equal(t, 1.0, testutil.ToFloat64(numPrograms))

	// it should decrement the programs_active metric when another Program is deleted
	deleteProgramCallback(nil)
	assert.Equal(t, 0.0, testutil.ToFloat64(numPrograms))

	// it should not decrement the programs_active metric when a Program is deleted and the metric is already at 0
	deleteProgramCallback(nil)
	assert.Equal(t, 0.0, testutil.ToFloat64(numPrograms))
}
```

operator/config/manager/manager.yaml
rquitales: PKO v1 was scaffolded with `operator-sdk` pre v1, which did not use `controller-runtime`/`kubebuilder` under the hood. It seems like the older version of the framework was modelled using `kube-state-metrics`  (where 2 ports were exposed, one for the operator, and another for metrics on the CRs itself, [ref](kube-state-metrics)). Kubebuilder uses a more simplified approach which exposes a globabl prometheus registry that we can add additional metrics to.

With the implementation of metrics being consolidated into 1 port, and industry standards moving to secure metrics by default, now seemed like a good time to make this change. WDYT?

go.mod
rquitales: Updated the `go.mod` file by manually deleting the `indirect` block and letting it regenerate.

operator/internal/controller/pulumi/metrics_program_test.go
rquitales: I thought about this, and in fact, I originally wrote this test using std Golang. However, I switched to using ginkgo ultimately, for consistency with other tests that we already have that test metrics. Also, this appeared to be more of a BDD style test, which is why I ultimately aligned with using Ginkgo.

Since we're already taking linting sacrifices in other files due to ginkgo, I'll leave this as is since I believe consistency here is more important.

blampe: Yeah totally agree one port is the most reasonable. AFAICT the other port wasn't actually used in v1. I'm still wondering if it's OK to change the port, though. Existing service monitors would stop scraping metrics, right?

operator/config/manager/manager.yaml
blampe: Yeah totally agree one port is the most reasonable. AFAICT the other port wasn't actually used in v1. I'm still wondering if it's OK to change the port, though. Existing service monitors would stop scraping metrics, right?

Issue #694, opened 2024-09-26
Option to retain or delete the workspace after successful sync


EronWright:
The stack controller provisions a workspace on demand, and it is fine to delete the workspace after sync was successful. Obviously the user is trading off performance/efficiency.

Similar to the "reclaim policy" of a PVC.
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaim-policy

For the retain case, we should consider making the workspace pods be ["burstable"](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#burstable) so that they consume minimal resources at rest but are allowed to consume unlimited memory resources during a deployment operation.

For example:
```
kind: Stack
spec:
  workspaceReclaimPolicy: retain|delete
```

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

terekete:
- scale to zero would be a great feature, not everyone cares about the amount of time it runs vs more the cost.

project0:
This is really required, i plan to spin up quiet a few stacks for ephemeral environments and it just eats unnecessary resources

PR #707, opened 2024-10-04
Set some default resource requests on the workspace pod

### LINKED ISSUE #698: Set some default resource requests on the workspace pod
Manager has limits on it already -- currently has guaranteed QoS.

Related to #694 and probably a pre-req -- set a small request limit to give the workspace pod burst-able QoS.

Additional considerations:
- use [GOMEMLIMIT](https://hsleep.medium.com/optimizing-memory-management-in-go-with-gomemlimit-da8ed23e2dcb) to tell the agent to not use more than the requested memory for itself. Consider setting it using the [Downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#downwardapi-resourceFieldRef). Avoid setting it on the child processes (use `SetMemoryLimit` in code?).
- check for zombie sub-processes.

EronWright:
### Proposed changes

Implements good defaults for the workspace resource, using a ["burstable"](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#burstable) approach.
Since a workspace pod's utilization is bursty - with low resource usage during idle times and with high resource usage during deployment ops - the pod requests a small amount of resources (64mb, 100m) to be able to idle. A deployment op is able to use much more memory - all available memory on the host.

Users may customize the resources (e.g. to apply different requests and/or limits).  For large/complex Pulumi apps, it might make sense to reserve more memory and/or use https://github.com/pulumi/pulumi-kubernetes-operator/issues/694.

The agent takes some pains to stay within the requested amount, using a programmatic form of the [GOMEMLIMIT](https://weaviate.io/blog/gomemlimit-a-game-changer-for-high-memory-applications) environment variable. The agent detects the requested amount via the Downward API.  We don't use `GOMEMLIMIT` to avoid propagating it to sub-processes, and because the format is a Kubernetes 'quantity'.

It was observed that zombies weren't being cleaned up, and this was leading to resource exhaustion. Fixed by using [tini](https://github.com/krallin/tini/) as the entrypoint process (PID 1).

### Related issues (optional)

Closes #698


operator/api/auto/v1alpha1/workspace_types.go
```diff
@@ -34,11 +34,6 @@ const (
 	SecurityProfileBaseline SecurityProfile = "baseline"
 	// SecurityProfileRestricted applies the restricted security profile.
 	SecurityProfileRestricted SecurityProfile = "restricted"
-
-	// SecurityProfileBaselineDefaultImage is the default image used when the security profile is 'baseline'.
-	SecurityProfileBaselineDefaultImage = "pulumi/pulumi:latest"
-	// SecurityProfileRestrictedDefaultImage is the default image used when the security profile is 'restricted'.
-	SecurityProfileRestrictedDefaultImage = "pulumi/pulumi:latest-nonroot"
```
EronWright, line 41: Rationale: moving these constants and the associated 'defaulting' logic to `webhook/auto/v1alpha1` where, in the future, a true webhook would apply the defaults eagerly.  This PR does NOT implement a webhook, it simply applies the defaults during reconciliation for simplicity.

operator/cmd/main.go
```diff
@@ -86,7 +86,7 @@ func main() {
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
 	flag.StringVar(&programFSAddr, "program-fs-addr", envOrDefault("PROGRAM_FS_ADDR", ":9090"),
 		"The address the static file server binds to.")
-	flag.StringVar(&programFSAdvAddr, "program-fs-adv-addr", envOrDefault("PROGRAM_FS_ADV_ADDR", ""),
+	flag.StringVar(&programFSAdvAddr, "program-fs-adv-addr", envOrDefault("PROGRAM_FS_ADV_ADDR", "localhost:9090"),
```
EronWright, line 89: Rationale: I was finding that the vscode launch configuration for the operator was failing because an advertised address is required.  This seems like a reasonable default.

operator/internal/controller/auto/workspace_controller.go
```diff
@@ -419,9 +445,10 @@ func newStatefulSet(ctx context.Context, w *autov1alpha1.Workspace, source *sour
 									ContainerPort: WorkspaceGrpcPort,
 								},
 							},
-							Env:     w.Spec.Env,
-							EnvFrom: w.Spec.EnvFrom,
-							Command: command,
+							Env:        env,
+							EnvFrom:    w.Spec.EnvFrom,
+							Command:    command,
+							WorkingDir: "/share/workspace",
```
EronWright, line 451: Rationale: It is a good time to include a small usability improvement, of setting the working directory of the container to the Pulumi program workspace. This automatically sets the initial directory when you shell into a pod using vscode.

operator/internal/webhook/auto/v1alpha1/workspace_webhook.go
```diff
@@ -0,0 +1,91 @@
+/*
+Copyright 2024 The Kubernetes authors.
```
rquitales, line 2: Should be Pulumi

operator/internal/webhook/auto/v1alpha1/workspace_webhook.go
```diff
@@ -0,0 +1,91 @@
...
+
+	if w.Spec.SecurityProfile == "" {
+		w.Spec.SecurityProfile = autov1alpha1.SecurityProfileRestricted
+	}
+
+	if w.Spec.Image == "" {
+		switch w.Spec.SecurityProfile {
+		case autov1alpha1.SecurityProfileRestricted:
+			w.Spec.Image = SecurityProfileRestrictedDefaultImage
+		default:
```
rquitales, line 73: This is a no-op. Should we default to a fall-through instead? ie. setting image as `SecurityProfileBaselineDefaultImage`

operator/api/auto/v1alpha1/workspace_types.go
EronWright: Let me say, the benefit of applying defaults eagerly (with a webhook) rather than lazily (during reconciliation) is stability; one may change the default later without affecting existing workloads.

blampe (approved):
Love that we killed the zombies!

My gut says the `SetMemoryLimit` call feels premature -- after all , if we OOM it'll probably be due to npm and not our little agent. But it probably doesn't hurt. We should definitely pull some profiles if we observe a leak.

operator/internal/webhook/auto/v1alpha1/workspace_webhook.go
```diff
@@ -0,0 +1,90 @@
...
+	// SecurityProfileRestrictedDefaultImage is the default image used when the security profile is 'restricted'.
+	SecurityProfileRestrictedDefaultImage = "pulumi/pulumi:latest-nonroot"
+)
+
+// // SetupWorkspaceWebhookWithManager registers the webhook for Workspace in the manager.
+// func SetupWorkspaceWebhookWithManager(mgr ctrl.Manager) error {
+// 	return ctrl.NewWebhookManagedBy(mgr).For(&autov1alpha1.Workspace{}).
+// 		WithDefaulter(&WorkspaceCustomDefaulter{}).
+// 		Complete()
+// }
```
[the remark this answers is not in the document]
EronWright: To be used later when we implement a webhook, which I felt was overkill for this PR.

EronWright:
@blampe here's the article that convinced me to give the system a hint that we're trying to stay within the 'requests'.
https://weaviate.io/blog/gomemlimit-a-game-changer-for-high-memory-applications

blampe:
> @blampe here's the article that convinced me to give the system a hint that we're trying to stay within the 'requests'. https://weaviate.io/blog/gomemlimit-a-game-changer-for-high-memory-applications

Right, rephrasing my earlier comment I don't think the agent falls into this high-memory category. It can run under 100MiB and handles one request at a time -- its heap should be pretty quiet :) Child processes will eat most of our memory, hence why it felt premature to me, but again it doesn't really matter.

PR #708, opened 2024-10-04
[V2] Clean up utils file

rquitales:
### Proposed changes

This PR removes the custom `min` and `max` functions for the `time.Duration` type. Go v1.21 introduced built-in functions so we can rely on them. This PR also adds a simple test case for the existing `exactlyOneOf` function.

### DIFF
--- a/operator/internal/controller/pulumi/utils.go
+++ b/operator/internal/controller/pulumi/utils.go
@@ -16,24 +16,6 @@ limitations under the License.
 
 package pulumi
 
-import (
-	"time"
-)
-
-func min(a, b time.Duration) time.Duration {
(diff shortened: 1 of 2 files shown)


EronWright (approved):
Were the min/max functions not being used in the stack controller?

rquitales:
> Were the min/max functions not being used in the stack controller?

`min` and `max` are used in the reconcile function to determine the requeue after time:
https://github.com/pulumi/pulumi-kubernetes-operator/blob/096b8462db56324347cbb04a967705b898d75824/operator/internal/controller/pulumi/stack_controller.go#L705-L713

PR #676, opened 2024-09-20
[v2] Capture stack outputs

blampe:
This returns stack outputs from the agent and records them in a secret.

* `OutputValue` is returned from the agent and contains raw JSON-encoded bytes for the output.
* Each `Update` owns a corresponding `-stack-outputs` secret.
* The secret includes a `pulumi.com/secrets` annotation with a list of sensitive fields, and the Stack API uses this to scrub outputs for the Stack's status.

PR #709, opened 2024-10-04
[v2] Retry failed updates with exponential backoff

### LINKED ISSUE #677: implement a back-off for re-attempts at a failed update
The stack controller seems to retry failed updates without using a backoff.  The built-in reconciliation backoff kicks in only when reconcile returns an error, which isn't the case here. 

Consider looking at `lastUpdate.endTime` to implement a backoff strategy when `lastUpdate.status` is `failed`. Assumedly the stack would stay marked as Reconciling (rather than Stalled).

blampe:
Currently, if our automation APIs call fail they return non-nil errors to the operator. In #676 I modified `Update` to translate these errors into a "failed" status on the Update/Stack, but other operations (preview etc.) still surface these errors and automatically re-queue.

We'd like to retry these failed updates much less aggressively than we retry transient network errors, for example. To accomplish this we do a few things:

* We consolidate the update controller's streaming logic for consistent error handling across all operations.
* We return errors with known gRPC status codes as-is, but unknown status codes are translated into failed results for all operations. 
* We start tracking the number of times a stack has attempted an update. This is used to determine how much exponential backoff to apply.
* A failed update is considered synced for a cooldown period before we retry it. The cooldown period starts at 5 minutes and doubles for every failed attempt, eventually maxing out at 24 hours.

Fixes #677


operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -679,12 +680,27 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
-		instance.Status.LastUpdate.State == shared.SucceededStackStateMessage &&
-		(isStackMarkedToBeDeleted ||
-			(instance.Status.LastUpdate.LastSuccessfulCommit == currentCommit &&
-				(!sess.stack.ContinueResyncOnCommitMatch || time.Since(instance.Status.LastUpdate.LastResyncTime.Time) < resyncFreq)))
+		((instance.Status.LastUpdate.State == shared.SucceededStackStateMessage &&
+			(isStackMarkedToBeDeleted ||
+				(instance.Status.LastUpdate.LastSuccessfulCommit == currentCommit &&
+					(!sess.stack.ContinueResyncOnCommitMatch || time.Since(instance.Status.LastUpdate.LastResyncTime.Time) < resyncFreq)))) ||
+			(!isStackMarkedToBeDeleted &&
+				instance.Status.LastUpdate.State == shared.FailedStackStateMessage && time.Since(instance.Status.LastUpdate.LastResyncTime.Time) < cooldown))
```
blampe, line 703: This is really hairy! I'll probably break this out and put a table test around it.

EronWright: Maybe some inline funcs to give names to the sub-expressions?

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -798,10 +819,14 @@ func (r *StackReconciler) markStackFailed(sess *StackReconcilerSession, instance
 		LastAttemptedCommit: current.Commit,
 		Permalink:           shared.Permalink(update.Status.Permalink),
 		LastResyncTime:      metav1.Now(),
+		Attempts:            1,
 	}
 	if last != nil {
 		instance.Status.LastUpdate.LastSuccessfulCommit = last.LastSuccessfulCommit
 	}
+	if last != nil && last.LastAttemptedCommit == current.Commit {
```
EronWright, line 827: May I suggest the following logic, to mean that we reset the atttempts whenever there's a generation change (as seen in big the `synced` expression).  Your expression wouldn't reset the cooldown in the face of a configuration property change.

```go
if last != nil && last.Generation == current.Generation {
   instance.Status.LastUpdate.Attempts += last.Attempts
}

File: operator/internal/controller/pulumi/stack_controller.go [hunk 3]
```diff
@@ -841,6 +867,10 @@ func (r *StackReconciler) markStackSucceeded(ctx context.Context, instance *pulu
 		LastSuccessfulCommit: current.Commit,
 		Permalink:            shared.Permalink(update.Status.Permalink),
 		LastResyncTime:       metav1.Now(),
+		Attempts:             int64(1),
+	}
+	if last != nil && last.LastAttemptedCommit == current.Commit {
+		instance.Status.LastUpdate.Attempts += last.Attempts
```
EronWright, line 873: I see, in the periodic resync, we'd see attempts increment at each run.  I think that is mildly confusing and is overkill.  I would keep it at 1.

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -1331,3 +1361,14 @@ func patchObject[T any, V any](base T, patch V) (*T, error) {
 
 	return &result, nil
 }
+
+func exactlyOneOf(these ...bool) bool {
```
EronWright, line 1365: We have this in utils.go already right?

EronWright (changes requested):

operator/api/pulumi/shared/stack_types.go
```diff
@@ -388,6 +388,10 @@ type StackUpdateState struct {
 	Permalink Permalink `json:"permalink,omitempty"`
 	// LastResyncTime contains a timestamp for the last time a resync of the stack took place.
 	LastResyncTime metav1.Time `json:"lastResyncTime,omitempty"`
+	// Attempts records how many times the update has been attempted. If the
+	// update failed it will be periodically retried with exponential backoff
+	// in case the failure was due to transient conditions.
+	Attempts int64 `json:"attempts"`
```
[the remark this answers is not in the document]
EronWright: Perhaps a field name of `Failures` would be more targeted and clear than `Attempts` where I wonder how to interpret it in some cases.

operator/internal/controller/pulumi/stack_controller.go
blampe: Yep that's a good point!

operator/internal/controller/pulumi/stack_controller.go
blampe: Jinxed with #708, but no need to keep that file around for a function that's only used in one place.

operator/internal/controller/pulumi/stack_controller.go
blampe: `!isStackMarkedToBeDeleted` was in here because we had a test that expected finalize to immediately retry a failed deletion, but we should do the same backoff in that case so I fixed the test.

I broke out `isSynced` into its own function to make it easier to unit test.

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -679,12 +680,27 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 	if resyncFreq.Seconds() < 60 {
 		resyncFreq = time.Duration(60) * time.Second
 	}
+
+	// If the update failed, allow it to be retried with an exponential
+	// cooldown. We start with a 5 minute cooldown and double that for each
+	// failed attempt, up to a max of 24 hours. Failed Updates are considered
```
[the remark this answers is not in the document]
blampe: Right now we only use this backoff when we receive "unknown" status codes, everything else (canceled, unavailable, etc.) will continue to get retried with the built-in reconcile backoff. We can get a lot more granular with this but I thought this strikes a good balance between short-lived blips and harder errors.

operator/internal/controller/auto/update_controller.go
```diff
@@ -579,3 +432,127 @@ func outputsToSecret(owner *autov1alpha1.Update, outputs map[string]*agentpb.Out
...
+	var res result
+
+	for {
+		stream, err := s.Recv()
+		if err == io.EOF {
+			break
+		}
+		if err != nil && status.Code(err) != codes.Unknown {
+			// Surface gRPC errors.
+			return nil, err
```
EronWright, line 474: What happens in this situation w.r.t. the status? Is the error assumed to be transient? 

You might get PermissionDenied for example, or a InvalidArgument, some may be retryable.

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -671,22 +672,7 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
-	}
-	synced := instance.Status.LastUpdate != nil &&
-		instance.Status.LastUpdate.Generation == instance.Generation &&
-		instance.Status.LastUpdate.State == shared.SucceededStackStateMessage &&
-		(isStackMarkedToBeDeleted ||
-			(instance.Status.LastUpdate.LastSuccessfulCommit == currentCommit &&
-				(!sess.stack.ContinueResyncOnCommitMatch || time.Since(instance.Status.LastUpdate.LastResyncTime.Time) < resyncFreq)))
-
-	if synced {
+	if isSynced(instance, currentCommit, isStackMarkedToBeDeleted) {
```
EronWright, line 675: nit: `isStackMarkedToBeDeleted` is redundant if you use `instance.DeletionTimestamp != nil`.

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -671,22 +672,7 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
 		// transition to ready, and requeue reconciliation as necessary to detect
 		// branch updates and resyncs.
 		instance.Status.MarkReadyCondition()
```
EronWright, line 678: If synced is true due to backoff, we shouldn't mark ourselves as ready, right? Maybe you'd want to check the lastUpdate's status.

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -842,12 +837,15 @@ func (r *StackReconciler) markStackSucceeded(ctx context.Context, instance *pulu
 		Permalink:            shared.Permalink(update.Status.Permalink),
 		LastResyncTime:       metav1.Now(),
 	}
+	if last != nil && last.Generation == current.Generation {
+		instance.Status.LastUpdate.Failures = last.Failures
+	}
```
EronWright, line 842: I would advocate for resettting the failure count to zero here. Once the update is successful, retaining the failure count seems debatable.  A couple of weird cases:
1. In the case of period resync, the failure count stays at non-zero in perpetuity.
2. In the case of sporatic errors (fail, success, fail), one might use backoff prematurely.

When a container recovers from a crashloop, does the status still reflect the situation or are "events" the only historical record at that point?

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -875,9 +873,72 @@ func newStackReconcilerSession(
...
+// succeeded.
+func cooldown(stack *pulumiv1.Stack) time.Duration {
+	cooldown := time.Duration(0)
+	if stack.Status.LastUpdate == nil {
+		return cooldown
+	}
+	if stack.Status.LastUpdate.State == shared.FailedStackStateMessage {
+		cooldown = 5 * time.Minute
+		cooldown *= time.Duration(math.Exp2(float64(stack.Status.LastUpdate.Failures)))
+		cooldown = min(24*time.Hour, cooldown)
```
EronWright, line 924: This seems not aggressive enough to cover short-term transient errors like a network connection error.

operator/internal/controller/pulumi/stack_controller_test.go
```diff
@@ -623,12 +711,27 @@ var _ = Describe("Stack Controller", func() {
...
-				_, err := reconcileF(ctx)
-				Expect(err).NotTo(HaveOccurred())
-				ByResyncing()
+			When("within cooldown period", func() {
+				It("backs off exponentially", func(ctx context.Context) {
+					res, err := reconcileF(ctx)
+					Expect(err).NotTo(HaveOccurred())
+					// 5 minutes * 2^2
+					Expect(res.RequeueAfter).To(BeNumerically("~", time.Duration(20*time.Minute), time.Minute))
+					ByMarkingAsReady()
```
EronWright, line 723: Why would the stack be ready when it is in backoff?  Consider, for example, the Stack dependencies feature, where one stack waits for another to be ready.  You wouldn't want to unblock a dependent.

operator/internal/controller/auto/update_controller.go
blampe: Transient errors will be handled with "fast" backoff, and non-transient errors will get this "slow" backoff.

I expanded slow backoff from just the Unknown status to now also Unauthenticated, PermissionDenied, and InvalidArgument.

Context cancelation and deadline exceeded will be handled with fast backoff.

LMK if that sounds reasonable.

operator/internal/controller/pulumi/stack_controller.go
blampe: Good point especially re: the dependencies feature. I changed this to mark as stalled if the update failed.

operator/internal/controller/pulumi/stack_controller.go
blampe: Lowered the initial backoff to 1 minute.

operator/internal/controller/pulumi/stack_controller.go
blampe: > In the case of sporatic errors (fail, success, fail), one might use backoff prematurely. Or is that intentional?

Very good point, this isn't intention. We should definitely reset here.

operator/internal/controller/pulumi/stack_controller.go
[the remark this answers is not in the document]
blampe: See latest comment here https://github.com/pulumi/pulumi-kubernetes-operator/pull/709#discussion_r1797448063

operator/internal/controller/pulumi/stack_controller_test.go
blampe: Changed to reconciling.

EronWright (approved):

Issue #704, opened 2024-10-03
Stop defaulting within the template blocks


EronWright:
The type of the `StackSpec.workspaceTemplate` field should be an "apply" variant, similar to [k8s.io/client-go/applyconfigurations](https://pkg.go.dev/k8s.io/client-go/applyconfigurations#section-directories); it should not set any default values.  Otherwise it makes for noise in the higher-level APIs and overstates their intentions.

For example, when we apply a given stack:
```yaml
kind: Stack
spec:
  workspaceTemplate:
    spec:
      image: pulumi/pulumi:3.134.1-nonroot
```

We get:

```yaml
kind: Stack
spec:
  workspaceTemplate:
    metadata: {}
    spec:
      image: pulumi/pulumi:3.134.1-nonroot
      resources: {}
      securityProfile: restricted
      serviceAccountName: default
```

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

blampe:
I think this is more to do with the WorkspaceSpec embedded in the Stack. We probably want an annotation to disable defaults on it (if there is such a thing) or create a separate Partial/Embedded type without defaults.

https://github.com/pulumi/pulumi-kubernetes-operator/blob/58b053960356ce484dee24a47f5aaed475ba9f7d/operator/config/crd/bases/pulumi.com_stacks.yaml#L9340-L9344

EronWright:
@blampe there's a code generator for just this purpose (applyconfigurations), to generate an EmbeddedWorkspaceSpec.

The same issue is present with PodTemplate in the WorkspaceSpec. IIRC there was a problem with DeepCopy being missing on the generated code, need to revisit it.

PR #713, opened 2024-10-09
[v2] Use an apply configuration for the workspace template

blampe:
This adds `applyconfiguration-gen` to generate apply configurations for our APIs. The workspace template is updated to use it.

It's a moderately uphill battle to get apply configurations working with controllers due to various pieces not including built-in support for them. In practice this means we need to cut a couple small corners -- in particular implementing our own `DeepCopy` and tweaking the generated `GroupVersion` code.

Fixes https://github.com/pulumi/pulumi-kubernetes-operator/issues/704.


operator/internal/controller/pulumi/session_test.go
```diff
@@ -357,17 +359,17 @@ func TestSetupWorkspace(t *testing.T) {
...
-					Spec: &autov1alpha1.WorkspaceSpec{
-						Image:              "custom-image",
-						ServiceAccountName: "custom-service-account",
-						PodTemplate: &autov1alpha1.EmbeddedPodTemplateSpec{
-							Metadata: autov1alpha1.EmbeddedObjectMeta{
+					Spec: &apply.WorkspaceSpecApplyConfiguration{
+						Image:              ptr.To("custom-image"),
+						ServiceAccountName: ptr.To("custom-service-account"),
+						PodTemplate: &apply.EmbeddedPodTemplateSpecApplyConfiguration{
+							Metadata: &apply.EmbeddedObjectMetaApplyConfiguration{
```
blampe, line 372: If we're happy with the direction of this I can take it further and also use an apply config for our pod template -- it should be very minor.

operator/internal/controller/pulumi/session_test.go
```diff
@@ -357,17 +359,17 @@ func TestSetupWorkspace(t *testing.T) {
 						},
 					},
 				},
-				WorkspaceTemplate: &autov1alpha1.EmbeddedWorkspaceTemplateSpec{
-					Metadata: autov1alpha1.EmbeddedObjectMeta{
+				WorkspaceTemplate: &shared.WorkspaceApplyConfiguration{
+					ObjectMetaApplyConfiguration: &metav1apply.ObjectMetaApplyConfiguration{
```
blampe, line 363: Worth calling out that `ObjectMetaApplyConfiguration` differs from `EmbeddedWorkspaceTemplateSpec` in that it does expose all of the underlying fields (like generation) even if they can't actually be set. I think it makes for a more bloated/confusing schema but I don't see a good way around it.

operator/internal/controller/pulumi/session_test.go
EronWright: That would be pretty wonderful.

EronWright (approved):
Awesome!  Good job finding a workaround for the DeepCopy issue.

operator/internal/controller/pulumi/session_test.go
rquitales: This approach is similar to `PodTemplateSpec` so it shouldn't add confusion to K8s users.

rquitales (approved):

operator/internal/controller/pulumi/session_test.go
blampe: I spoke too soon -- this is actually quite involved. I'll put up a WIP with what I have so far.

Issue #650, opened 2024-09-06
Optimize stack configuration step


EronWright:
The Workspace controller is expected to configure the stack with the supplied configuration values.  The current implementation calls the `SetAllConfig` RPC method for each config item, and this is proving to be slow.  Would be nice to set in bulk, but be considerate about the ordering of the config elements.

Also, consider enhancing the Workspace config block to natively understand configmap/secret refs.  Today, the Stack controller handles references via general extensibility, i.e. via the pod template. A first-class option would be better.

cleverguy25:
Added to epic https://github.com/pulumi/pulumi-kubernetes-operator/issues/586

PR #718, opened 2024-10-16
[v2] Set config all at once

blampe:
We currently issue one `SetAllConfig` RPC for each user-specified config. This is slow but it has important correctness guarantees:

1. The order we apply config matters -- if the user specifies `foo: foo` followed by `foo: bar`, the net result must always be `foo: bar`.
2. The Pulumi CLI (and therefore Automation API) only allows specifying `--path` on an all-or-nothing basis. This is bad for us because we potentially have a blend of path and non-path keys.

Ideally we would be able to supply all of our configs to the automation API in a single call, and in the case where _all_ of our config keys are path-like (or all are not path-like) we actually can do that because we no longer have limitation (2).

This PR makes that possible in the general case by transforming our config keys in a way that allows us to treat them as if they are all path-like.

In particular:
* The agent's `SetAllConfig` handler is modified to take a list of configs instead of a map in order to preserve config order. The top-level `path` param is also removed and handled on a per-key basis.
* While resolving configs, we escape any non-path keys so subsequent path parsing treats them as verbatim. For example `foo.bar` gets escaped as `["foo.bar"]`.
* We can then supply all of our keys at once to Automation API with `Path: true`.
* If there are no configs to set then the operator doesn't invoke `SetAllConfig`.

Fixes https://github.com/pulumi/pulumi-kubernetes-operator/issues/650


EronWright (approved):

blampe:
Confirmed config still propagates as expected. Of course the StacK API is currently just a map with no path information, so all of the path optimizations are only relevant for the Workspace API. However it does give users the ability -- although somewhat clunky -- to specify somewhat structured (`path: true`) config via the `workspaceTemplate`.

PR #723, opened 2024-10-17
Add the serviceAccountName field to Stack.spec

### LINKED ISSUE #720: Re-add the serviceAccountName field to Stack.spec
Virtually all stacks should have an associated service account, so the serviceAccountName should be a top-level field as opposed to being buried in the workspaceTemplate (which is intended for extensibility and advanced use cases).

EronWright:
### Proposed changes

Adds a top-level `serviceAccountName` to the Stack spec, since it is practically a required element for every stack.
Note that the `workspaceTemplate` would take precedence.
The Workspace applies the default value of `default` if none is specified.

Side-effect: some outdated manifests were regenerated.

### Related issues (optional)

Closes #720

### DIFF
--- a/operator/config/manager/manager.yaml
+++ b/operator/config/manager/manager.yaml
@@ -44,7 +44,7 @@ spec:
             - --leader-elect
             - --health-probe-bind-address=:8081
             - --program-fs-adv-addr=pulumi-kubernetes-operator.$(POD_NAMESPACE).svc.cluster.local:80
-            - --zap-log-level=error
+            - --zap-log-level=info
             - --zap-time-encoding=iso8601
           ports:
           - containerPort: 8383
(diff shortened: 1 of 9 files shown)


operator/config/manager/manager.yaml
```diff
@@ -44,7 +44,7 @@ spec:
             - --leader-elect
             - --health-probe-bind-address=:8081
             - --program-fs-adv-addr=pulumi-kubernetes-operator.$(POD_NAMESPACE).svc.cluster.local:80
-            - --zap-log-level=error
+            - --zap-log-level=info
```
EronWright, line 47: Slipped in this fix, is now consistent with Helm chart and is more useful at info level.

blampe (approved):

PR #724, opened 2024-10-18
Update the "max reconciles" for each controller to 25

EronWright:
### Proposed changes

Use 25 by default to facilitate the blocking operations in the auto API, and support the `MAX_CONCURRENT_RECONCILES` environment variable.

### DIFF
--- a/operator/cmd/main.go
+++ b/operator/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strconv"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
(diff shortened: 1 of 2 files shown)


operator/cmd/main.go
```diff
@@ -145,9 +147,17 @@ func main() {
 		metricsServerOptions.FilterProvider = filters.WithAuthenticationAndAuthorization
 	}
 
+	controllerOpts := config.Controller{
+		MaxConcurrentReconciles: 25,
+	}
+	if s, ok := os.LookupEnv("MAX_CONCURRENT_RECONCILES"); ok {
+		controllerOpts.MaxConcurrentReconciles, _ = strconv.Atoi(s)
```
blampe, line 154: Should probably panic if this doesn't parse, otherwise it'll be zero which I'm guessing would do... nothing?

blampe (commented):
👍

PR #745, opened 2024-11-08
support for ReconcileRequest annotation

### LINKED ISSUE #398: Force a re-run even when there is no change in the git repository
## Hello!

## Issue details

Feedback from on of our customers:

> It would be great if we can have a documented and reliable way to force a pulumi run even when a git change has not be made.

Customer is implementing currently `continueResyncOnCommitMatch` but would still love to see this feature in the future.

### LINKED ISSUE #672: Support for pulumi.com/reconciliation-request
One of the features of `Stack` is support for the Flux on-demand reconciliation protocol ([ref](https://pkg.go.dev/github.com/fluxcd/pkg/runtime/predicates#ReconcileRequestedPredicate)), based on `pulumi.com/reconciliation-request` annotation.

The intention is to force a stack update when the stack is in a ready state, independent of the branch polling and periodic resyncs. As an annotation, the stack generation does not change. The workspace needn't be replaced, which notably happens whenever generation does change.  The last update should incorporate the annotation value.

The "prerequisites" feature leverages this feature when the `SucceededWithinDuration` field on a given prerequisite is set.  The field is proactive in nature; for example, given that stack A depends on stack B, a value of 5m on A will force a resync of stack B every 5 minutes even if B doesn't use periodic resync. A more "passive" (assumedly incorrect) interpretation of the feature, would be to check that B was recently synced while relying only on the natural causes of resync. See also: https://github.com/pulumi/pulumi-kubernetes-operator/pull/443

```go
	// SucceededWithinDuration gives a duration within which the prerequisite must have reached a
	// succeeded state; e.g., "1h" means "the prerequisite must be successful, and have become so in
	// the last hour". Fields (should there ever be more than one) are not intended to be mutually
	// exclusive.
	SucceededWithinDuration *metav1.Duration `json:"succeededWithinDuration,omitempty"`
```

EronWright:
### Proposed changes

Implements support for requesting a re-sync of the stack (i.e. `pulumi up`) using the
`pulumi.com/reconciliation-request` annotation.

The logic of checking the prerequisites was moved to be after the check for "is synced", since the prereq check has the side-effect of touching the parent stack, and that shouldn't happen unless the child is needing a sync.  

### Related issues (optional)

Closes #672 
Closes #398

### DIFF
--- a/operator/api/pulumi/shared/stack_types.go
+++ b/operator/api/pulumi/shared/stack_types.go
@@ -381,6 +381,8 @@ type StackOutputs map[string]apiextensionsv1.JSON
 type StackUpdateState struct {
 	// Generation is the stack generation associated with the update.
 	Generation int64 `json:"generation,omitempty"`
+	// ReconcileRequest is the stack reconcile request associated with the update.
+	ReconcileRequest string `json:"reconcileRequest,omitempty"`
 	// Name is the name of the update object.
 	Name string `json:"name,omitempty"`
 	// Type is the type of update.
@@ -398,7 +400,7 @@ type StackUpdateState struct {
 	// Failures records how many times the update has been attempted and
 	// failed. Failed updates are periodically retried with exponential backoff
 	// in case the failure was due to transient conditions.
-	Failures int64 `json:"failures"`
+	Failures int64 `json:"failures,omitempty"`
 }
 
 // StackUpdateStatus is the status code for the result of a Stack Update run.
@@ -435,6 +437,8 @@ type Permalink string
 type CurrentStackUpdate struct {
 	// Generation is the stack generation associated with the update.
 	Generation int64 `json:"generation,omitempty"`
+	// ReconcileRequest is the stack reconcile request associated with the update.
+	ReconcileRequest string `json:"reconcileRequest,omitempty"`
 	// Name is the name of the update object.
 	Name string `json:"name,omitempty"`
 	// Commit is the commit SHA of the planned update.
--- a/operator/internal/controller/pulumi/stack_controller.go
+++ b/operator/internal/controller/pulumi/stack_controller.go
@@ -68,11 +68,12 @@ var (
 )
 
 type errRequirementOutOfDate struct {
-	LastResyncTime time.Time
+	LastUpdateName string
+	TimeElapsed    time.Duration
 }
 
 func (e errRequirementOutOfDate) Error() string {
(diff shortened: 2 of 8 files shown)


operator/api/pulumi/shared/stack_types.go
```diff
@@ -398,7 +400,7 @@ type StackUpdateState struct {
 	// Failures records how many times the update has been attempted and
 	// failed. Failed updates are periodically retried with exponential backoff
 	// in case the failure was due to transient conditions.
-	Failures int64 `json:"failures"`
+	Failures int64 `json:"failures,omitempty"`
```
blampe, line 403: 👍

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -584,22 +637,22 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
-				// world next time around.
-				return reconcile.Result{}, fmt.Errorf("annotating prerequisite to force requeue: %w", err)
+
+			var errOutOfDate *errRequirementOutOfDate
+			if errors.As(requireErr, &errOutOfDate) {
+				// annotate the out-of-date stack so that it'll be resynced.
+				// The value is arbtrary but here may be understood as the time of the resync that was deemed out-of-date.
+				// Such a value is idempotent (don't want to spam the underlying stack with new requests)
+				// and supports having multiple dependent stacks.
+				setReconcileRequestAnnotation(prereqStack, errOutOfDate.LastResyncTime.Format(time.RFC3339))
```
blampe, line 647: Do we want to no-op if it already had the annotation?

blampe (approved):

operator/internal/controller/pulumi/stack_controller.go
EronWright: If the annotation already has the expected value, then the Update call would effectively be a no-up.

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -865,7 +920,7 @@ func (r *StackReconciler) markStackFailed(sess *stackReconcilerSession, instance
 	if last != nil {
 		instance.Status.LastUpdate.LastSuccessfulCommit = last.LastSuccessfulCommit
 	}
-	if last != nil && last.Generation == current.Generation {
+	if last != nil && last.Generation == current.Generation && last.ReconcileRequest == current.ReconcileRequest {
```
EronWright, line 923: Rationale: reset the backoff when a resync is requested.

PR #746, opened 2024-11-10
Some stack outputs cannot be saved to Secret

### LINKED ISSUE #743: Some stack outputs cannot be saved to Secret
### What happened?

Stack outputs have a free-form name, for example an output might be named `foo bar` (i.e. has whitespace). Such outputs cannot be stored into the `Secret` that is associated with the `Update` object, and the controller sees an error from the API Server to the effect that the secret cannot contain a key with an invalid name.

```
data[foo bar]: Invalid value: "foo bar": a valid config key must consist of 
alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', 
regex used for validation is '[-._a-zA-Z0-9]+')
```

### Example

Make a simple program that emits an output with an abnormal name.

```yaml/
name: example
runtime: yaml
outputs:
  "foo bar": "baz"
```

### Output of `pulumi about`

```plain
CLI          
Version      3.132.0
Go Version   go1.23.1
Go Compiler  gc

Plugins
KIND      NAME  VERSION
language  yaml  unknown

Host     
OS       darwin
Version  14.5
Arch     arm64

```

### Additional context

The solution is assumedly to sanitize the name to conform with the restriction.  An alternative would be to pack all the outputs into a single key, but that would make the Secret be less useful.

### Contributing

EronWright:
### Proposed changes

This PR sanitizes the output names from a given `Update` when the outputs are stored into a `Secret`.

### Related issues (optional)

Closes #743

### DIFF
--- a/operator/internal/controller/auto/update_controller.go
+++ b/operator/internal/controller/auto/update_controller.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -485,8 +486,13 @@ func outputsToSecret(owner *autov1alpha1.Update, outputs map[string]*agentpb.Out
 	}})
 
 	secrets := []string{}
-	for k, v := range outputs {
-		// v.Value is already JSON-encoded bytes,
+	for outputName, v := range outputs {
+		// note: v.Value is already JSON-encoded bytes
+		k := outputName
+		if strings.Contains(k, " ") {
+			// sanitize the outputName to be a valid secret key
+			k = strings.ReplaceAll(k, " ", "_")
+		}
 		s.Data[k] = v.Value
 		if v.Secret {
 			secrets = append(secrets, k)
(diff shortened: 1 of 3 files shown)


operator/internal/controller/auto/update_controller.go
```diff
@@ -485,8 +486,13 @@ func outputsToSecret(owner *autov1alpha1.Update, outputs map[string]*agentpb.Out
...
 
 	secrets := []string{}
-	for k, v := range outputs {
-		// v.Value is already JSON-encoded bytes,
+	for outputName, v := range outputs {
+		// note: v.Value is already JSON-encoded bytes
+		k := outputName
+		if strings.Contains(k, " ") {
+			// sanitize the outputName to be a valid secret key
+			k = strings.ReplaceAll(k, " ", "_")
```
EronWright, line 494: I intend to use the official regex for what a Secret may accept.

blampe (approved):
Stamping to unblock - 👍 to using the official regex.

Issue #762, opened 2024-11-26
New commits not detected


blampe:
### What happened?

The operator doesn't currently requeue in cases where we're using git sources. This means we don't detect new commits.

### Example

...

### Output of `pulumi about`

...

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #763, opened 2024-11-26
Fix requeue for git commit tracking

blampe:
Fixes a bug where we were accidentally cancelling a requeue when we intended to poll again after a minute.

Fixes https://github.com/pulumi/pulumi-kubernetes-operator/issues/762.

### DIFF
--- a/operator/internal/controller/pulumi/stack_controller.go
+++ b/operator/internal/controller/pulumi/stack_controller.go
@@ -770,7 +770,11 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 				// Reconcile every resyncFreq to check for new commits to the branch.
 				pollFreq := resyncFreq(instance)
 				log.Info("Commit hash unchanged. Will poll for new commits.", "pollFrequency", pollFreq)
-				requeueAfter = max(1*time.Second, min(pollFreq, requeueAfter))
+				if requeueAfter > 0 {
+					requeueAfter = max(1*time.Second, min(pollFreq, requeueAfter))
+				} else {
+					requeueAfter = max(1*time.Second, pollFreq)
+				}
 			} else {
 				log.Info("Commit hash unchanged.")
 			}
--- a/operator/internal/controller/pulumi/stack_controller.go
+++ b/operator/internal/controller/pulumi/stack_controller.go
@@ -752,20 +752,28 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 		}
 
 		// Requeue reconciliation as necessary to detect branch updates and
-		// resyncs.
-
+		// resyncs. The logic finds the smallest requeue time (if any) among various polling loops.
 		requeueAfter := time.Duration(0)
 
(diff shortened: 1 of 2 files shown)


operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -756,22 +756,16 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
+		if sess.stack.ContinueResyncOnCommitMatch || stack.Branch != "" {
 			requeueAfter = max(1*time.Second, time.Until(instance.Status.LastUpdate.LastResyncTime.Add(resyncFreq(instance))))
 		}
 		if stack.GitSource != nil {
-			trackBranch := len(stack.GitSource.Branch) > 0
-			if trackBranch {
-				// Reconcile every resyncFreq to check for new commits to the branch.
-				pollFreq := resyncFreq(instance)
-				log.Info("Commit hash unchanged. Will poll for new commits.", "pollFrequency", pollFreq)
-				requeueAfter = min(requeueAfter, pollFreq)
```
blampe, line 771: `requeueAfter` defaults to 0 so this `min` was always disabling the requeue.

EronWright: Good find, that's totally the problem.

Regarding the duplication of code here, I think my rationale is that the resync frequency is unduly be coupled between branch tracking and the periodic resync, ie. the `resyncFrequencySeconds`. I anticipate new git-specific fields for branch tracking, such that one could say "check the branch every minute, and do a re-sync (for drift detection) every hour".

EronWright (approved):

operator/internal/controller/pulumi/stack_controller.go
EronWright: I took the liberty of reverting the optimization, for clarity.

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -752,23 +752,25 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
+		// resyncs. The logic finds the smallest requeue time (if any) among various polling loops.
 		requeueAfter := time.Duration(0)
 
+		// Try again with exponential backoff if the update failed.
 		if instance.Status.LastUpdate.State == shared.FailedStackStateMessage {
 			requeueAfter = max(1*time.Second, time.Until(instance.Status.LastUpdate.LastResyncTime.Add(cooldown(instance))))
 		}
-		if sess.stack.ContinueResyncOnCommitMatch {
+		// Schedule another poll if ContinueResyncOnCommitMatch is set, for drift detection or to maintain dynamic resources.
+		if instance.Status.LastUpdate.State == shared.SucceededStackStateMessage && sess.stack.ContinueResyncOnCommitMatch {
```
EronWright, line 763: The "ContinueResyncOnCommitMatch" polling loop should activate only when the last update was successful.  If failed, the backoff loop should prevail (and is likely a shorter interval).

PR #802, opened 2025-02-04
Use 'parallel' policy for workspace pod rollouts to avoid stalls

### LINKED ISSUE #801: Workspace pod may stall during rollout
### What happened?

When the workspace spec is undeployable, e.g. due to an invalid docker image, the rollout fails.  Unfortunately, attempts to update the spec aren't effective at unblocking the system, due to a limitation of StatefulSet:

https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#forced-rollback

One must manually delete the workspace pod to unblock the system.

See also:
- https://github.com/kubernetes/kubernetes/issues/67250#issuecomment-414754525
- https://github.com/kubernetes/kubernetes/issues/78007

### Example

Lots of ways to trigger this:
- an invalid pod specification (e.g. invalid image)
- unable to fetch the program source (e.g. from git), leading to pod startup failure

### Output of `pulumi about`

PKO v2.0.0-beta.3

### Additional context

Let's keep in mind some requirements.
1. The system should respect the pod termination grace period when updating or deleting the pod, to gracefully cancel any in-flight Pulumi operation.
2. The system should preserve the persistent volume(s) of the workspace pod, in case the user makes use of PVs.
3. The system should wipe the temporary directory within the pod.

A possible solution may be to use the "parallel" pod management strategy:
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#parallel-pod-management

### Contributing

EronWright:
### Proposed changes

This PR seeks to address this issue ([k8s: "Forced rollback"](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#forced-rollback)) that occurs when the workspace pod is in a crashloop:
> When using [Rolling Updates](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#rolling-updates) with the default [Pod Management Policy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies) (OrderedReady), it's possible to get into a broken state that requires manual intervention to repair.

The `parallel` policy seems to enable the statefulset controller to forcibly remove a pod when a new revision is available. The controller seems to obey the termination grace period as is important, and I can't think of any other negatives.  But there's a concern in the k8s community about this approach: https://github.com/kubernetes/website/issues/47085

Note that a workspace consists of one replica, and is rather like a singleton with good behavior w.r.t. Pulumi state locking and compatible with persistent volumes.

### Related issues (optional)

Closes #801

### DIFF
--- a/operator/e2e/e2e_test.go
+++ b/operator/e2e/e2e_test.go
@@ -118,6 +118,10 @@ func TestE2E(t *testing.T) {
 					assert.NoError(t, err)
 					t.Log(string(out))
 				}
+
+				// cleanup
+				cmd = exec.Command("kubectl", "delete", "-f", "e2e/testdata/random-yaml-nonroot")
+				require.NoError(t, run(cmd))
 			},
 		},
 		{
@@ -135,6 +139,10 @@ func TestE2E(t *testing.T) {
 
 				assert.Equal(t, `"[secret]"`, string(stack.Status.Outputs["secretOutput"].Raw))
 				assert.Equal(t, `"foo"`, string(stack.Status.Outputs["simpleOutput"].Raw))
+
+				// cleanup
+				cmd = exec.Command("bash", "-c", "envsubst < e2e/testdata/git-auth-nonroot/* | kubectl delete -f -")
+				require.NoError(t, run(cmd))
 			},
 		},
 		{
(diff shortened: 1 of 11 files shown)


rquitales:
Per offline discussion, the workaround makes sense, but we should somehow exercise this in an e2e test to ensure that taking advantage of the implementation detail of `parallel` for StatefulSet updates carries through for future Kubernetes releases.

Ref from docs:

> This option only affects the behaviour for scaling operations. Updates are not affected.

operator/e2e/e2e_test.go
```diff
@@ -75,7 +77,6 @@ func TestE2E(t *testing.T) {
 		{
 			name: "random-yaml-nonroot",
 			f: func(t *testing.T) {
-				t.Parallel()
```
EronWright, line 78: I will remove these from this PR, and open a separate PR together with some other e2e changes to support k3d.

operator/internal/controller/auto/workspace_controller.go
```diff
@@ -187,6 +188,14 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 	err = r.Patch(ctx, ss, client.Apply, client.FieldOwner(FieldManager))
 	if err != nil {
+		// issue-801 - migration logic for 2.0-beta.3 to 2.0.0
```
rquitales, line 191: nit:
```suggestion
		// issue-801 - migration logic for 2.0.0-beta.3 to 2.0.0
```

operator/e2e/e2e_test.go
rquitales: We could also use something like gotestfmt in this case to make parallel outputs nicer.

rquitales (approved):

PR #804, opened 2025-02-05
feat: add `WorkspaceReclaimPolicy` to Stack spec

### LINKED ISSUE #694: Option to retain or delete the workspace after successful sync

rquitales:
### Proposed Changes  

This PR introduces the `WorkspaceReclaimPolicy` option in the Stack spec. By default, it is set to `Retain`, preserving the current behavior where a Workspace (along with its associated StatefulSet and pod) is not deleted after a successful sync. When set to `Delete`, the Workspace pod is removed after a successful sync, effectively scaling the workload down to zero.  

#### Testing  

- Added envtests to verify that the Workspace object is deleted.  
- Manually validated on a GKE cluster.  

### Related Issues (Optional)  

Fixes #694

### DIFF
--- a/operator/internal/controller/pulumi/stack_controller.go
+++ b/operator/internal/controller/pulumi/stack_controller.go
@@ -802,6 +802,15 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 			log.Info("Not commit tracking. Will wait for resync.")
 		}
 
+		// Delete the workspace if the reclaim policy is set to delete.
+		if instance.Spec.WorkspaceReclaimPolicy == shared.WorkspaceReclaimDelete {
+			log.Info("Deleting workspace as reclaim policy is set to delete")
+			err := sess.DeleteWorkspace(ctx)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+
 		return reconcile.Result{RequeueAfter: requeueAfter}, saveStatus()
 	}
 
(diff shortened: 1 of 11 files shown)


rquitales:

- #804 ◀
    - #805

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -1359,6 +1368,16 @@ func (sess *stackReconcilerSession) CreateWorkspace(ctx context.Context) error {
 	return nil
 }
 
+// DeleteWorkspace deletes the workspace object. This is manually called when a Stack is successfully synced,
+// and the Stack has a WorkspaceReclaimPolicy of Delete.
+func (sess *stackReconcilerSession) DeleteWorkspace(ctx context.Context) error {
+	if err := sess.kubeClient.Delete(ctx, sess.ws); err != nil {
```
EronWright, line 1374: Check for apierrors.IsNotFound

EronWright (approved):

Issue #806, opened 2025-02-06
Update CR failure message is not surfaced to Stack CR object


rquitales:
### What happened?

When an Update operation fails, the message/reason for failure is not surfaced back to the Stack CR object. This makes it hard for users to understand why a stack operation failed without diving into the logs, or the statuses of any Update CRs.

This occurs because we do not set the high level summary field for `Update.Status.Message`.

https://github.com/pulumi/pulumi-kubernetes-operator/blob/673ab77245d95556f5c9cb184c9024f384f72db3/operator/internal/controller/auto/update_controller.go#L288-L307

### Example

Current Stack status output which gives no details about why an update failed:

```
status:
  conditions:
  - lastTransitionTime: "2025-02-06T00:11:09Z"
    message: reconciliation is in progress
    reason: NotReadyInProgress
    status: "False"
    type: Ready
  - lastTransitionTime: "2025-02-06T00:11:09Z"
    message: 2 update failure(s)
    reason: RetryingAfterFailure
    status: "True"
    type: Reconciling
  lastUpdate:
    failures: 2
    generation: 4
    lastAttemptedCommit: sha256:f335a9e0bc445b0dbe3187371f56017bcdd66e23b68c6eda54910eeb48d5e3a0
    lastResyncTime: "2025-02-06T00:14:27Z"
    lastSuccessfulCommit: sha256:f335a9e0bc445b0dbe3187371f56017bcdd66e23b68c6eda54910eeb48d5e3a0
    name: nginx-stack-194d89b9f33
    state: failed
    type: up
  observedGeneration: 4
  outputs:
    availableReplicas: 1
```

### Output of `pulumi about`

-

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Issue #736, opened 2024-11-01
Report status when stack is locked


EronWright:
A given stack may become stuck due to locking.  Can the operator reflect this condition to the user? Perhaps the update could have a Reason for its Failed condition, that would help the Stack controller produce a better status.

Here's an example of how the problem may occur.

```
 pulumi 2024-10-03T00:08:30.369Z INFO    pulumi  Updating (dev)
 pulumi 2024-10-03T00:08:30.370Z INFO    pulumi
 pulumi 2024-10-03T00:08:45.917Z WARN    pulumi  error: [500] Internal Server Error
 pulumi 2024-10-03T00:08:45.995Z ERROR   server  up completed with an error      {"error": "failed to run update: exit status 255\ncode: 255\nstdout: Updating (dev)\n\n\nstderr: error: [500] Internal Server Error\n\n"}
 pulumi 2024-10-03T00:08:45.996Z ERROR   cmd.serve.grpc  finished streaming call with code Unknown       {"grpc.start_time": "2024-10-03T00:08:28Z", "system": "grpc", "span.kind": "server", "grpc.service": "agent.AutomationService", "grpc.method": "Up", "peer.address": "10.1.33.119:44098", "error": "failed to run update: exit status 255\ncode: 255\nstdout: Updating (dev)\n\n\nstderr: error: [500] Internal Server Error\n\n", "grpc.code": "Unknown", "grpc.time_ms": 17067.732}
random-yaml-workspace-0 pulumi 2024-10-03T00:08:46.868Z INFO    server  selected a stack        {"name": "dev"}
random-yaml-workspace-0 pulumi 2024-10-03T00:08:48.256Z INFO    cmd.serve.grpc  finished unary call with code OK        {"grpc.start_time": "2024-10-03T00:08:46Z", "system": "grpc", "span.kind": "server", "grpc.service": "agent.AutomationService", "grpc.method": "SelectStack", "peer.address": "10.1.33.119:44478", "grpc.code": "OK", "grpc.time_ms": 2078.996}
 pulumi 2024-10-03T00:08:49.541Z INFO    pulumi  Updating (dev)
 pulumi 2024-10-03T00:08:49.542Z INFO    pulumi
 pulumi 2024-10-03T00:08:49.895Z WARN    pulumi  error: [409] another update is currently in progress or was interrupted. Try again later.) Try again later
[...]

PR #805, opened 2025-02-05
fix: do not delete workspace pod on authz errors

rquitales:
### Proposed Changes  

This PR updates the workspace controller to run the grpc `WhoAmI` command before proceeding with reconciliation. This change helps surface authentication issues early in the process.  

If authentication fails, the workspace pod is retained rather than deleted, as it remains in a pristine state. This avoids unnecessary pod recreation and reduces the time required to spin up a new StatefulSet pod. Additionally, it prevents excessive workspace pod churn in the event of persistent authentication failures.  

#### Example Log Output on Authentication Failure  

```shell
2025-02-05T18:37:33.029Z	INFO	Applying StatefulSet	{"controller": "workspace-controller", "controllerGroup": "auto.pulumi.com", "controllerKind": "Workspace", "Workspace": {"name":"nginx-k8s-stack","namespace":"pulumi-kubernetes-operator"}, "namespace": "pulumi-kubernetes-operator", "name": "nginx-k8s-stack", "reconcileID": "bca83f73-73aa-4004-bce0-14aa8c4bf48a", "revision": "112271", "hash": "ec3e2de049a62cdb9a133f4c141df826", "source": {"Generation":1,"ForceRequest":"","Git":{"URL":"https://github.com/pulumi/examples","Ref":"4ce6027439cc8c9cc60f5704abc7d2204a07d98e","Dir":"/kubernetes-ts-nginx","Shallow":false,"SSHPrivateKey":null,"Username":null,"Password":null,"Token":null},"Flux":null,"Local":null}}
2025-02-05T18:37:33.151Z	INFO	Connecting to workspace pod	{"controller": "workspace-controller", "controllerGroup": "auto.pulumi.com", "controllerKind": "Workspace", "Workspace": {"name":"nginx-k8s-stack","namespace":"pulumi-kubernetes-operator"}, "namespace": "pulumi-kubernetes-operator", "name": "nginx-k8s-stack", "reconcileID": "bca83f73-73aa-4004-bce0-14aa8c4bf48a", "revision": "112271", "addr": "nginx-k8s-stack-workspace.pulumi-kubernetes-operator:50051"}
[...]

PR #807, opened 2025-02-06
feat: report status when stack is locked

rquitales:
### Proposed changes

- **Surfaces Locked Stack Errors:**  
  - The agent server now returns a structured response instead of an error when the Pulumi CLI returns a `409` (Conflict) error.  
  - Clients can now determine if a stack is locked without parsing error streams.  
  - Introduced a new boolean response field, `isStackLocked`, for easy client-side detection.  

- **Improves Stack CR Status Updates:**  
  - The error message from `UpdateCR.status.message` is now correctly propagated to `StackCR.status.lastUpdate.message`.  
  - Ensures locked stack errors are surfaced in the Stack CR's status subresource.  

#### Example Stack CR Status Block  
```
status:
    conditions:
    - lastTransitionTime: "2025-02-06T00:11:09Z"
      message: reconciliation is in progress
      reason: NotReadyInProgress
      status: "False"
      type: Ready
    - lastTransitionTime: "2025-02-06T00:11:09Z"
      message: 4 update failure(s)
      reason: RetryingAfterFailure
      status: "True"
      type: Reconciling
    lastUpdate:
      failures: 4
      generation: 4
      lastAttemptedCommit: sha256:f335a9e0bc445b0dbe3187371f56017bcdd66e23b68c6eda54910eeb48d5e3a0
      lastResyncTime: "2025-02-06T00:26:33Z"
      lastSuccessfulCommit: sha256:f335a9e0bc445b0dbe3187371f56017bcdd66e23b68c6eda54910eeb48d5e3a0
      message: Another update is currently in progress
      name: nginx-stack-194d8a6b139
      state: failed
      type: up
    observedGeneration: 4
    outputs:
      availableReplicas: 1
```

#### Example Update CR Status Block  
```
status:
  conditions:
  - lastTransitionTime: "2025-02-06T00:26:33Z"
    message: ""
    observedGeneration: 1
    reason: Complete
    status: "False"
    type: Progressing
  - lastTransitionTime: "2025-02-06T00:26:33Z"
    message: Another update is currently in progress
    observedGeneration: 1
    reason: StackLocked
    status: "True"
    type: Failed
  - lastTransitionTime: "2025-02-06T00:26:33Z"
    message: ""
    observedGeneration: 1
    reason: Updated
    status: "True"
    type: Complete
  endTime: "1970-01-01T00:00:00Z"
  message: Another update is currently in progress
  observedGeneration: 1
  startTime: "1970-01-01T00:00:00Z"
```

### Testing  

- Added envtests to validate that statuses are correctly surfaced.  
- Manually validated on a GKE cluster.  

### Related Issues  

Fixes: #806  
Fixes: #736


rquitales:

- #804
    - #805
        - #807 ◀

agent/pkg/server/server.go
```diff
@@ -390,18 +390,24 @@ func (s *Server) Preview(in *pb.PreviewRequest, srv pb.AutomationService_Preview
...
 	res, err := stack.Preview(ctx, opts...)
 	if err != nil {
-		s.log.Errorw("preview completed with an error", zap.Error(err))
-		return err
+		// If the error occurs due to a locked stacked, we need to surface this to the client.
+		if !strings.Contains(err.Error(), "Conflict: Another update is currently in progress") {
+			s.log.Errorw("preview completed with an error", zap.Error(err))
+			return err
+		}
+		stackLocked = true
```
EronWright, line 401: The `res` variable is not really usable if the preview failed, so we should return here.  Can we produce an error struct with an error code like `Locked`, that the client can unmarshal?

agent/pkg/proto/agent.proto
```diff
@@ -143,6 +143,7 @@ message RefreshResult {
   string stderr = 2;
   UpdateSummary summary = 3;
   optional string permalink = 4;
+  optional bool is_stack_locked = 5;
```
EronWright, line 146: I would definitely prefer an error structure of some kind, and probably nor embedded into the result struct.  Maybe the method can return a oneof?

operator/internal/controller/auto/update_controller.go
```diff
@@ -294,6 +297,18 @@ func (rs *reconcileSession) updateStatus(ctx context.Context, obj *autov1alpha1.
...
+	// Order of precedence: failed > complete > progressing
+	switch {
+	case rs.failed.Status == metav1.ConditionTrue:
+		obj.Status.Message = rs.failed.Message
+	case rs.complete.Status == metav1.ConditionTrue:
+		obj.Status.Message = rs.complete.Message
+	case rs.progressing.Status == metav1.ConditionTrue:
+		obj.Status.Message = rs.progressing.Message
+	}
+
```
EronWright, line 311: Is it really necessary to compute `obj.Status.Message` at this time, in updateStatus?  In `Result()` is the logic that updates the message; isn't `s.obj.Status.Message = res.GetSummary().Message` enough?

operator/internal/controller/auto/update_controller.go
```diff
@@ -621,6 +641,7 @@ type getResulter[T stream] struct {
 type result interface {
 	GetSummary() *agentpb.UpdateSummary
 	GetPermalink() string
+	GetIsStackLocked() bool
```
EronWright, line 644: Would prefer not to see `GetIsStackLocked` here.

The result interface is, I believe, intended to resemble the auto api.  I imagine that if the auto api had support for the "stack is locked" issue, it would be presented as an `err` with a status code or similar.  Can we make the GRPC interface be similar to that?

EronWright (changes requested):

operator/internal/controller/auto/update_controller.go
```diff
@@ -332,6 +334,11 @@ func (u *reconcileSession) Preview(ctx context.Context, obj *autov1alpha1.Update
 	reader := streamReader[agentpb.PreviewStream]{receiver: res, l: l, u: u, obj: obj}
 	_, err = reader.Result()
 	if err != nil {
+		// Update the status/conditions.
+		if err := u.updateStatus(ctx, obj); err != nil {
```
rquitales, line 338: Previously, when we get a grpc error, we do not surface this back to the Update CR's status/conditions. This ensures we always update the conditions/status.

operator/internal/controller/auto/update_controller.go
```diff
@@ -612,6 +634,48 @@ func (s streamReader[T]) Result() (result, error) {
 	return res, fmt.Errorf("didn't receive a result")
 }
 
+// setStatusFromGRPCErr sets the Update object status blocks based on the result of a gRPC error.
```
rquitales, line 637: This is the core logic for determining if we have a structured error, and uses it for surfacing information about Pulumi errors back to the user.

agent/pkg/server/pulumi_errors.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"strings"
+
+	pb "github.com/pulumi/pulumi-kubernetes-operator/v2/agent/pkg/proto"
+	"google.golang.org/grpc/status"
+)
+
+type knownPulumiErrors map[string]*pb.PulumiErrorInfo
+
+// knownErrors is a map where the keys are string snippets of the CLI error message,
```
rquitales, line 26: We can add more known Pulumi CLI errors here to surface back to the controller.

operator/internal/controller/auto/update_controller.go
rquitales: Second iteration of this removes this logic here.

operator/internal/controller/auto/update_controller.go
```diff
@@ -612,6 +634,48 @@ func (s streamReader[T]) Result() (result, error) {
...
+// If the error contains a structured PulumiErrorInfo, we'll use that information to surface the error.
+func (s streamReader[T]) setStatusFromGRPCErr(grpcErr error) error {
```
EronWright, line 639: It seems like this function doesn't really seek to change the error that is returned to the controller, and I'm not sure it should anyway. It seems to mutate the status, similarly to the block in the "Update complete" case.  So, maybe `setStatusFromGRPCErr` should not have a return value, for clarity.  It is purely a procedure.

operator/internal/controller/pulumi/stack_controller_test.go
```diff
@@ -537,6 +537,34 @@ var _ = Describe("Stack Controller", func() {
 				Expect(r.Recorder.(*record.FakeRecorder).Events).To(Receive(matchEvent(pulumiv1.StackUpdateFailure)))
 			})
 
+			When("a stack is locked", func() {
```
EronWright, line 540: Could you add a case for the generic error situation? ('Reason: Unknown')

EronWright (changes requested):
Looking really good

operator/internal/controller/auto/update_controller.go
rquitales: Ah, yes, we should just return the high-level grpc error anyway, since it'll contain more information.

rquitales: Updated to only update the status block.

EronWright (approved):
LGTM

PR #810, opened 2025-02-11
Garbage collection for Update objects

### LINKED ISSUE #733: Garbage collection for Update objects
`Update` objects accumulate indefinitely in the system, and the operator should cleanup the old completed updates, e.g. after a week or so.

Should the Update controller handle the cleanup uniformly?  Or should the Stack controller handle the cleanup for the updates that it owns?  I would lean towards the former, by introducing a `Update.spec.ttlSecondsAfterFinished` field that the `Stack` controller may set and is implemented by the Update controller.

Verify that the output secret for a given update is cleaned up.

See https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/

EronWright:
### Proposed changes

Implements a retention policy for completed Update objects, based on a simple TTL with the default value of 24h. Update objects that are in a completed state and are older than the TTL are summarily deleted.

#### New API: `Update.spec.ttlAfterCompleted` 

A new field to enable garbage collection of this update, once completed. Optional; no default value.

#### Update Controller

Updated to implement the cleanup logic. When a TTL is set on a completed update, the controller looks to the last transition time of the Completed status condition. If there's time remaining, the controller requeues the object for later, otherwise it deletes the update object.

#### Stack Controller

Updated to assign a 24-hour TTL on the update objects that it creates. A future enhancement would be to provide an `updateTemplate` field to allow for further control. A user may edit an existing Update to change the TTL.

#### Update Controller Tests

A new unit test was added to exercise the retention logic.

### Related issues (optional)

Closes #733


deploy/yaml/install.yaml
```diff
@@ -19552,7 +19577,11 @@ spec:
                   TargetDependents allows updating of dependent targets discovered but not
                   specified in the Target list
                 type: boolean
+              ttlAfterCompleted:
+                description: TTL for a completed update object.
+                type: string
               type:
+                description: Type of the update to perform.
```
thomas11, line 19584: The description doesn't seem right?

I think it should include what format is expected, and how to disable GC (if possible).

operator/internal/controller/auto/update_controller_test.go
```diff
@@ -46,56 +46,85 @@ import (
...
-			By("Reconciling the created resource")
-			controllerReconciler := &UpdateReconciler{
-				Client: k8sClient,
-				Scheme: k8sClient.Scheme(),
-			}
+		// refresh the object and find its status condition(s)
+		obj = &autov1alpha1.Update{}
+		_ = k8sClient.Get(ctx, objName, obj)
+
+		// progressing = meta.FindStatusCondition(obj.Status.Conditions, autov1alpha1.UpdateConditionTypeProgressing)
```
thomas11, line 86: [nit] cleanup? Also line 53

thomas11 (approved):
LGTM as someone who isn't familiar with PKO.

deploy/yaml/install.yaml
rquitales: The descriptions are autogenerated from: https://github.com/pulumi/pulumi-kubernetes-operator/pull/810/files#diff-c203c1c46b5c663e63a1eeb95dacb02f899c2808c38f8b4645ee87bec34f6831.

In this case, `type` is actually a field describing what type of Update is being applied.

operator/api/auto/v1alpha1/update_types.go
```diff
@@ -39,10 +39,13 @@ const (
...
-	StackName     string `json:"stackName,omitempty"`
-
+	// Specify the Pulumi stack to select for the update.
+	StackName string `json:"stackName,omitempty"`
+	// Type of the update to perform.
 	Type UpdateType `json:"type,omitempty"`
-
+	// TTL for a completed update object.
+	// +optional
+	TtlAfterCompleted *metav1.Duration `json:"ttlAfterCompleted,omitempty"`
```
rquitales, line 48: Should we align naming to be similar to Jobs? `.ttlSecondsAfterFinished` Otherwise, we should be more specific about the accepted units of time here. Is it seconds or hours etc

EronWright: It is a metav1.Duration, not an integer.  I took a bit of time to look at newer APIs from kube SIGs, and they seem to prefer that the fields be like this, as opposed to the older style seen in Job.

I could use "finished" rather than "completed", but note that we have a `Completed` status condition.

rquitales: I agree with using `Completed`, as that's the nomenclature we're using for conditions. I'm more familiar with ttl fields using `int32` as their values, but I'm not opposed to using `time.Duration` instead.

PR #815, opened 2025-02-12
Use correct logic for OwnerReferencesChangedPredicate

EronWright:
### Proposed changes

Fixes the logic to correctly trigger reconciliation when the owner references have changed.


EronWright:
We unfortunately do not have tests for the watch logic because we haven't developed a technique for doing it.

PR #816, opened 2025-02-12
Use audience-scoped access token

### LINKED ISSUE #715: Use audience-scoped access token
The operator transmits its own service account token to the workspace pod for authentication purposes.  There is a risk of privilege elevation if the workspace pod were to capture the token and use it to impersonate the operator for arbitrary API Server interactions.

Kubernetes has a feature to prevent this: audience-scoped tokens. Rather than using the default token, the operator should call the `TokenRequest` API to get a token with a workspace-specific audience (e.g. the service endpoint).  The token then has a specific workspace as its audience, and is useless for any other purpose.  The `TokenReview` API validates such tokens as normal.  On the agent side, one attaches the expected audience to the context when calling `AuthenticateToken`:

```go
import "k8s.io/apiserver/pkg/authentication/authenticator"
...
ctx = authenticator.WithAudiences(ctx, []string{"http://my-audience"})
res, ok, err := a.authn.AuthenticateToken(ctx, token)
```

See also: https://adil.medium.com/how-to-use-tokenrequest-api-and-tokenvolume-projection-in-kubernetes-f007135b9994

EronWright:
### Proposed changes

This PR changes how the operator authenticates to the agent's RPC endpoint. Before this PR, the operator would pass (as a bearer token) its own service account token (the contents of `/var/run/secrets/kubernetes.io/serviceaccount/token`) to the agent. The problem is that a malicious agent could capture that token, and replay it to impersonate the operator in other calls to the API server.

The solution implemented in this PR is for the operator to use an audience-specific token, by setting the `aud` claim of the token to a workspace-specific value, and by removing the API Server as an audience.  The token for a workspace named `default/random-yaml` now resembles:

```json
{
  "aud": [
    "random-yaml.default"
  ],
  "iss": "https://kubernetes.default.svc.cluster.local",
  "sub": "system:serviceaccount:pulumi-kubernetes-operator:controller-manager"
}
```

The [TokenRequest API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) is the basis for this solution.  It produces short-lived tokens representing a given service account and intended for an arbitrary audience(s), signed by the API Server. The operator makes a token request for its own service account, whenever it needs to access a given workspace, and then attaches the resultant token to outgoing requests.

The "audience" value is arbitrary, often an https: URL.  This PR settles on the simple form `<name>.<namespace>` representing a given workspace.  e.g. `random-yaml.default`.  If we prefer to use a more literal service URL, it could be `grpc://random-yaml-workspace.default:50051` or similar. Note how the service endpoint has a `-workspace` suffix.

#### Agent: Audience Configuration and Validation

The agent accepts a new command-line flag `--kube-audience` accepting an array of values. It is a required parameter when `--auth-mode=kube`.

Incoming tokens are validated to have at least one matching audience.  The agent returns a status code of [UNAUTHENTICATED](https://grpc.io/docs/guides/status-codes/) when the audience is mismatched.

#### Operator: Workspace Spec Changes

The operator configures the workspace pod to have a workspace-specific audience (via the `--kube-audience` flag). Note that the format of the audience value is encapsulated in the operator.

#### Operator: Connection Manager and Token Cache

It is important to cache the tokens across reconciliation passes. Since multiple controllers connect to the workspace pod, the controller manager now provides a shared connection manager to encapsulate this.

The connection manager maintains an LRU cache of _token sources_ (keyed by workspace). When a controller requests a connection to a given workspace, the manager vends a token source to produce a series of fresh tokens.  The manager periodically (every 5 minutes) scrubs the cache of token sources that haven't been used in at least 2 hours, with the rationale that tokens generally have a 1-hour expiration.

Each token source maintains a cache of the latest token, and rotates the token when it is within 5 minutes of its expiration time.

#### Operator: New environment variable: POD_SA_NAME

The operator must know its own service account name, to be able to call into the TokenRequest API. The deployment scripts are adjusted to set the `POD_SA_NAME` environment variable using the Downward API.

When running the operator in an IDE, one should (at minimum) deploy the operator's SA (`pulumi-kubernetes-operator/controller-manager`). The user's own credentials are then naturally used to impersonate the SA, and there's no longer a need to have a local token file.

#### Operator: RBAC changes

The operator needs new permissions to be able to call the TokenRequest API to impersonate its own service account. A new `Role` and corresponding `RoleBinding` is introduced for this purpose. Note that the permission is not cluster-wide and is scoped to a specific `resourceName`.

### Testing

New/updated unit tests: 
- connection_test.go
- auth_test.go

### TODOs

- [x] Verify that the RBAC permissions of the operator's service account do allow for calls to the TokenRequest API.
- [x] Write unit tests for new behavior

### Examples

Here's a full example token:

```json
{
  "aud": [
    "random-yaml.default"
  ],
  "exp": 1739306657,
  "iat": 1739303057,
  "iss": "https://kubernetes.default.svc.cluster.local",
  "jti": "190a47cf-8963-44b4-b77c-18ece58bc4af",
  "kubernetes.io": {
    "namespace": "pulumi-kubernetes-operator",
    "serviceaccount": {
      "name": "controller-manager",
      "uid": "f48ad504-695a-4775-ba52-d66cd7ffdbb8"
    }
  },
  "nbf": 1739303057,
  "sub": "system:serviceaccount:pulumi-kubernetes-operator:controller-manager"
}
```

### Related issues (optional)

Closes #715


agent/pkg/client/credentials.go
```diff
@@ -0,0 +1,244 @@
+/*
+Copyright © 2024 Pulumi Corporation
```
rquitales, line 2: Incorrect year

rquitales (commented):
First pass of this PR looks good in general. Will take a deeper look at agent/pkg/client/credentials.go on the next pass.

agent/pkg/client/credentials.go
EronWright: True, and I suppose your second pass on linting will fix it.

EronWright:
Note that portions of credentials.go are adapted from these sources:
- https://github.com/kubernetes/client-go/blob/bf05f9c594a5707d9ade1ec2122847e66c1a149f/transport/token_source.go#L150-L204
- https://github.com/grpc/grpc-go/blob/0003b4fa356eb8a3aeef5572b6a998ad1207c274/credentials/oauth/oauth.go#L40-L58

agent/pkg/client/credentials_test.go
```diff
@@ -0,0 +1,362 @@
...
+	token      string
+	expiration time.Time
+	err        error
+
+	calls              int
+	serviceAccountName string
+	audiences          []string
+}
+
+var _ serviceAccountInterface = &mockServiceAccountInterface{}
```
rquitales, line 44: nit: we don't need this line since we're using `mockServiceAccountInterface` as `serviceAccountInterface` later L67, so incorrectly implementation of the interface will be surfaced anyway.

rquitales (approved):
Nice! Left a suggestion to improve test, but not blocking.

agent/pkg/client/credentials_test.go
```diff
@@ -0,0 +1,362 @@
...
+		{
+			name:    "service account not found",
+			err:     apierrors.NewNotFound(schema.GroupResource{Group: "", Resource: "serviceaccounts"}, "test"),
+			wantErr: true,
+		},
+		{
+			name:    "token resource forbidden",
+			err:     apierrors.NewForbidden(schema.GroupResource{Group: "", Resource: "serviceaccounts/token"}, "test", errors.New("testing")),
+			wantErr: true,
+		},
```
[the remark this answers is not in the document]
EronWright: It is true that the implementation code doesn't differentiate these cases, but I wanted to capture the main error cases.

agent/pkg/client/credentials_test.go
EronWright: I always make these declarations immediately before the implementation methods of the given interface.

Issue #823, opened 2025-02-13
Increase logging with helm chart


rquitales:
We should allow changing the logging level of the operator and agent using our Helm installation mode.

Stretch goal: 
- enable pulumi logging
- consider adding a `logging` field

See suggestion:
https://github.com/pulumi/pulumi-kubernetes-operator/issues/786#issuecomment-2621913592

rquitales:
Customizing the logging level of the operator via Helm should already be supported currently by setting the `controller.logLevel` value

Ref: https://github.com/pulumi/pulumi-kubernetes-operator/tree/master/deploy/helm/pulumi-operator#values

PR #824, opened 2025-02-13
Enable setting the log verbosity level of Pulumi CLI for a stack/workspace

rquitales:
### Proposed changes

This PR adds a new optional Stack spec field: `PulumiLogVerbosity`. This field determines the log verbosity level of pulumi up/down/refresh operations.

### Related issues (optional)

Closes: #823

### DIFF
--- a/agent/cmd/serve.go
+++ b/agent/cmd/serve.go
@@ -51,6 +51,7 @@ var (
 	_audiences          []string
 	_workspaceNamespace string
 	_workspaceName      string
+	_pulumiLogLevel     uint
 )
 
 // serveCmd represents the serve command
@@ -156,7 +157,8 @@ var serveCmd = &cobra.Command{
 
 		// Create the automation service
 		autoServer, err := server.NewServer(ctx, workspace, &server.Options{
-			StackName: _stack,
+			StackName:      _stack,
+			PulumiLogLevel: _pulumiLogLevel,
 		})
 		if err != nil {
 			return fmt.Errorf("unable to make an automation server: %w", err)
@@ -218,4 +220,5 @@ func init() {
 	serveCmd.Flags().StringArrayVar(&_audiences, "kube-audience", nil, "The audience to expect in the token (for kubernetes auth mode)")
 	serveCmd.Flags().StringVar(&_workspaceNamespace, "kube-workspace-namespace", os.Getenv("WORKSPACE_NAMESPACE"), "The Workspace object namespace (for kubernetes auth mode)")
 	serveCmd.Flags().StringVar(&_workspaceName, "kube-workspace-name", os.Getenv("WORKSPACE_NAME"), "The Workspace object name (for kubernetes auth mode)")
+	serveCmd.Flags().UintVar(&_pulumiLogLevel, "pulumi-log-level", 0, "The level of logging to use for the Pulumi CLI")
 }
--- a/operator/api/auto/v1alpha1/workspace_types.go
+++ b/operator/api/auto/v1alpha1/workspace_types.go
@@ -101,6 +101,13 @@ type WorkspaceSpec struct {
 	// +optional
 	PodTemplate *EmbeddedPodTemplateSpec `json:"podTemplate,omitempty"`
 
+	// PulumiLogVerbosity is the log verbosity level to use for the Pulumi CLI.
+	// If unset,verbose logging is disabled.
+	// See: https://www.pulumi.com/docs/iac/support/troubleshooting/#verbose-logging
+	// for more information about log levels.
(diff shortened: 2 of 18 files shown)


operator/api/auto/v1alpha1/workspace_types.go
```diff
@@ -101,6 +101,13 @@ type WorkspaceSpec struct {
 	// +optional
 	PodTemplate *EmbeddedPodTemplateSpec `json:"podTemplate,omitempty"`
 
+	// PulumiLogVerbosity is the log verbosity level to use for the Pulumi CLI.
+	// If unset,verbose logging is disabled.
+	// See: https://www.pulumi.com/docs/iac/support/troubleshooting/#verbose-logging
+	// for more information about log levels.
+	// +optional
+	PulumiLogVerbosity uint32 `json:"pulumiLogLevel,omitempty"`
```
EronWright, line 109: I think i would advocate for `logLevel` rather than `pulumiLogLevel` because the workspace is very much pulumi-specific.

operator/api/pulumi/shared/stack_types.go
```diff
@@ -159,6 +159,13 @@ type StackSpec struct {
 	// The default behavior is to retain the workspace. Valid values are one of "Retain" or "Delete".
 	// +optional
 	WorkspaceReclaimPolicy WorkspaceReclaimPolicy `json:"workspaceReclaimPolicy,omitempty"`
+
+	// (optional) PulumiLogVerbosity is the log verbosity level to use for the Pulumi CLI.
+	// If unset,verbose logging is disabled.
+	// See: https://www.pulumi.com/docs/iac/support/troubleshooting/#verbose-logging
+	// for more information about log levels.
+	// +optional
+	PulumiLogVerbosity uint32 `json:"pulumiLogLevel,omitempty"`
```
EronWright, line 168: I think we resisted adding first-class fields when it simply copies what's possible in template.  Let's just leave it at the template level, WDYT?

agent/cmd/serve.go
```diff
@@ -218,4 +220,5 @@ func init() {
 	serveCmd.Flags().StringArrayVar(&_audiences, "kube-audience", nil, "The audience to expect in the token (for kubernetes auth mode)")
 	serveCmd.Flags().StringVar(&_workspaceNamespace, "kube-workspace-namespace", os.Getenv("WORKSPACE_NAMESPACE"), "The Workspace object namespace (for kubernetes auth mode)")
 	serveCmd.Flags().StringVar(&_workspaceName, "kube-workspace-name", os.Getenv("WORKSPACE_NAME"), "The Workspace object name (for kubernetes auth mode)")
+	serveCmd.Flags().UintVar(&_pulumiLogLevel, "pulumi-log-level", 0, "The level of logging to use for the Pulumi CLI")
```
EronWright, line 223: Seems the default should be `1` to match the default log level in Pulumi.  I know it works with zero but one seems clearer to me.  WDYT?

EronWright (commented):
Looking good but I would pare it back a bit and use the `logLevel` field name.

agent/cmd/serve.go
rquitales: The default if `--verbose` is not provided is `0` [[ref](https://github.com/pulumi/pulumi/blob/425a5d703969ee9218b56e655b9085bc86909d1b/pkg/cmd/pulumi/pulumi.go#L331-L332)].

It differentiates between whether verbose logging was not enabled or not: https://github.com/pulumi/pulumi/blob/master/sdk/go/common/util/logging/log.go#L106. This suggests that there is a difference between `0` and `1`.

operator/api/pulumi/shared/stack_types.go
rquitales: It does make it easier to manage. Will update.

rquitales: That's fair. We can always add this top level field in the future. Always easier to add, than remove :)

operator/api/auto/v1alpha1/workspace_types.go
rquitales: True, but we also have verbosity for the agent/server itself. I'd rather have more specific names to avoid ambiguity.

rquitales: Note: the agent cmd supports `verbose` currently enables/disables verbose logging for the agent binary. Having `logLevel` instead of `pulumiLogLevel` may confuse users into thinking that the logLevel is for the agent binary itself, instead of the Pulumi CLI.

EronWright (approved):
Documentation change?

Issue #826, opened 2025-02-13
Don't reclaim workspace during backoff


EronWright:
When a stack update fails, the stack enters a backoff state. I think it desirable to retain the workspace in that case, e.g. for manual troubleshooting.  Note how `isSynced` may return true even the previous update was a failure.

If we really think that the workspace should be reclaimed during backoff, I would suggest that we change the policy API to cover each case.  Similar to statefulset.persistentVolumeClaimRetentionPolicy ([ref](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention)).
```yaml
workspaceReclaimPolicy:
  whenSuccessful: delete
  whenFailed: retain
```

PR #828, opened 2025-02-14
fix: do not delete the workspace pod when the last update failed

rquitales:
### Proposed changes

This PR fixes it so that the Workspace pod is not reclaimed when the update failed. We should only delete the pod if the sync was successful, and the update was also successful.

### Related issues (optional)

Closes: #826

### DIFF
--- a/deploy/crds/pulumi.com_stacks.yaml
+++ b/deploy/crds/pulumi.com_stacks.yaml
@@ -773,7 +773,8 @@ spec:
                 type: boolean
               workspaceReclaimPolicy:
                 description: |-
-                  WorkspaceReclaimPolicy specifies whether the workspace should be deleted or retained after the Stack is synced.
+                  WorkspaceReclaimPolicy specifies whether the workspace should be deleted or retained after the Stack is
+                  successfully synced, and the last update is in a succeeded state.
                   The default behavior is to retain the workspace. Valid values are one of "Retain" or "Delete".
                 enum:
                 - Retain
@@ -10409,7 +10410,8 @@ spec:
                 type: boolean
               workspaceReclaimPolicy:
                 description: |-
-                  WorkspaceReclaimPolicy specifies whether the workspace should be deleted or retained after the Stack is synced.
+                  WorkspaceReclaimPolicy specifies whether the workspace should be deleted or retained after the Stack is
+                  successfully synced, and the last update is in a succeeded state.
                   The default behavior is to retain the workspace. Valid values are one of "Retain" or "Delete".
                 enum:
                 - Retain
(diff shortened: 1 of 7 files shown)


operator/internal/controller/pulumi/stack_controller_test.go
```diff
@@ -939,6 +953,21 @@ var _ = Describe("Stack Controller", func() {
...
+							It("reconciles and delete the workspace pod", func(ctx context.Context) {
+								result, err := reconcileF(ctx)
+								Expect(err).NotTo(HaveOccurred())
+								ByMarkingAsReady()
+								By("not requeuing")
+								Expect(result).To(Equal(reconcile.Result{}))
+								By("deleting the Workspace object")
+								Expect(ws.GetName()).To(BeEmpty())
+							})
+						})
```
EronWright, line 970: I don't understand why this case is here, in a section that deals with 'ContinueResyncOnCommitMatch'.  I think this block assumes a successful last update.

operator/internal/controller/pulumi/stack_controller_test.go
```diff
@@ -788,6 +788,20 @@ var _ = Describe("Stack Controller", func() {
...
+
+				When("the WorkspaceReclaimPolicy is set to Delete", func() {
+					BeforeEach(func(ctx context.Context) {
+						obj.Spec.WorkspaceReclaimPolicy = shared.WorkspaceReclaimDelete
+					})
+					It("does not delete the workspace pod", func(ctx context.Context) {
+						result, err := reconcileF(ctx)
+						Expect(err).NotTo(HaveOccurred())
+						// 1 minute * 2^3
+						Expect(result.RequeueAfter).To(BeNumerically("~", 8*time.Minute, time.Minute))
```
EronWright, line 800: nit: I would omit these two lines because they aren't germane to the test, which seeks to show that it "does not delete the workspace pod".

deploy/crds/pulumi.com_stacks.yaml
```diff
@@ -773,7 +773,8 @@ spec:
                 type: boolean
               workspaceReclaimPolicy:
                 description: |-
-                  WorkspaceReclaimPolicy specifies whether the workspace should be deleted or retained after the Stack is synced.
+                  WorkspaceReclaimPolicy specifies whether the workspace should be deleted or retained after the Stack is
+                  successfully synced, and the last update is in a succeeded state.
```
EronWright, line 777: I'm struggling with the phrase "successfully synced".  May I suggest "WorkspaceReclaimPolicy specifies whether the workspace should be deleted or retained after the Stack is successfully updated".

operator/internal/controller/pulumi/stack_controller_test.go
rquitales: This is nested with the following contexts: `the last update was successful` > `at latest commit` > `ContinueResyncOnCommitMatch` > `a resync is due`. This ensures that we also delete the Workspace pod even if `ContinueResyncOnCommitMatch` is set to true, and the update was successful.

EronWright (approved):

PR #829, opened 2025-02-14
Use pulumi.com labels

### LINKED ISSUE #819: change the 'app' metadata used by operator
Stop using `app` Kubernetes labels for the statefulset/service/pod selector.  Use a PKO specific label.

EronWright:
### Proposed changes

This PR switches over to a different labeling strategy for:
- the child resources of workspaces (statefulsets, services, pods, secrets), and
- the child resources of stacks (workspaces, updates)

The rationale here is to avoid using the app.kubernetes.io labels because they have various meanings, the user might also be using them, and may trigger unwanted behavior like object tracking in argocd.  PKO uses labels for a functional purpose, e.g. in the selectors.

Regarding the specific labels, this PR seeks to relate the labels the apigroup that is using them.  The workspace controller uses labels in the `auto.pulumi.com` API group. The stack controller uses labels in the `pulumi.com` API group.

#### Workspace

Before:
```yaml
apiVersion: auto.pulumi.com/v1alpha1
kind: Workspace
metadata:
  labels:
    app.kubernetes.io/component: stack
    app.kubernetes.io/instance: random-yaml
    app.kubernetes.io/managed-by: pulumi-kubernetes-operator
    app.kubernetes.io/name: pulumi
```

After:
```yaml
apiVersion: auto.pulumi.com/v1alpha1
kind: Workspace
metadata:
  labels:
    pulumi.com/component: stack             # <--- the workspace is a part of the stack
    pulumi.com/stack-name: random-yaml
```

Note that this change causes replacement of the statefulset, due to some recovery logic that we have in the workspace controller.

#### Update

Labels weren't previously applied.

After:
```yaml
apiVersion: auto.pulumi.com/v1alpha1
kind: Update
metadata:
  labels:
    pulumi.com/component: stack       # <--- the update is a part of the stack
    pulumi.com/stack-name: random-yaml
```

#### Service

Before:
```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/component: workspace
    app.kubernetes.io/instance: random-yaml
    app.kubernetes.io/managed-by: pulumi-kubernetes-operator
    app.kubernetes.io/name: pulumi
```

After:
```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    auto.pulumi.com/component: workspace               # <--- the service is a part of the workspace
    auto.pulumi.com/workspace-name: random-yaml
```

Similar changes for the StatefulSet, for the Pod, and for the Secret made by each Update.

### Related issues (optional)

Closes #819

### DIFF
--- a/operator/internal/controller/auto/update_controller.go
+++ b/operator/internal/controller/auto/update_controller.go
@@ -547,6 +547,7 @@ func outputsToSecret(owner *autov1alpha1.Update, outputs map[string]*agentpb.Out
 		Name:       owner.Name,
 		UID:        owner.UID,
 	}})
+	s.Labels = labelsForUpdateSecret(owner)
 
 	secrets := []string{}
 	for outputName, v := range outputs {
@@ -569,6 +570,13 @@ func outputsToSecret(owner *autov1alpha1.Update, outputs map[string]*agentpb.Out
 	})
 
 	return s, nil
+}
+
+func labelsForUpdateSecret(u *autov1alpha1.Update) map[string]string {
+	return map[string]string{
+		"auto.pulumi.com/component":   "update",
+		"auto.pulumi.com/update-name": u.Name,
+	}
 }
 
 // stream is an interface constraint for the response streams consumable by a
--- a/operator/internal/controller/auto/update_controller.go
+++ b/operator/internal/controller/auto/update_controller.go
@@ -574,8 +574,8 @@ func outputsToSecret(owner *autov1alpha1.Update, outputs map[string]*agentpb.Out
 
 func labelsForUpdateSecret(u *autov1alpha1.Update) map[string]string {
 	return map[string]string{
-		ComponentAnnotation:     UpdateComponent,
-		WorkspaceNameAnnotation: u.Name,
+		ComponentLabel:  UpdateComponent,
+		UpdateNameLabel: u.Name,
 	}
 }
 
--- a/operator/internal/controller/auto/update_controller.go
+++ b/operator/internal/controller/auto/update_controller.go
@@ -574,8 +574,8 @@ func outputsToSecret(owner *autov1alpha1.Update, outputs map[string]*agentpb.Out
 
 func labelsForUpdateSecret(u *autov1alpha1.Update) map[string]string {
 	return map[string]string{
-		autov1alpha1.ComponentAnnotation:     autov1alpha1.UpdateComponent,
-		autov1alpha1.WorkspaceNameAnnotation: u.Name,
+		ComponentAnnotation:     UpdateComponent,
+		WorkspaceNameAnnotation: u.Name,
 	}
 }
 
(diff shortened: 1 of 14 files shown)


operator/internal/controller/auto/workspace_controller.go
```diff
@@ -433,10 +433,8 @@ func fqdnForService(w *autov1alpha1.Workspace) string {
 
 func labelsForStatefulSet(w *autov1alpha1.Workspace) map[string]string {
 	return map[string]string{
-		"app.kubernetes.io/name":       "pulumi",
-		"app.kubernetes.io/component":  "workspace",
-		"app.kubernetes.io/instance":   w.Name,
-		"app.kubernetes.io/managed-by": "pulumi-kubernetes-operator",
```
rquitales, line 439: Is there a rationale for removing the well known `app.kubernetes.io/managed-by` label? This makes it obvious to other tools what is managing these child objects.

operator/internal/controller/auto/update_controller.go
```diff
@@ -571,6 +572,13 @@ func outputsToSecret(owner *autov1alpha1.Update, outputs map[string]*agentpb.Out
 	return s, nil
 }
 
+func labelsForUpdateSecret(u *autov1alpha1.Update) map[string]string {
+	return map[string]string{
+		"auto.pulumi.com/component":   "update",
```
rquitales, line 577: Can we start creating constants for these label keys in a common package?

operator/internal/controller/auto/workspace_controller.go
EronWright: I was hesitant to use a subset of app labels, seemed a bit weird.  The definition of the `app.kubernetes.io/managed-by` label is:
> The tool being used to manage the operation of an application.

To me, it identifies the tool that is servicing the application at large.  Since PKO doesn't know the application to which the workload is associated, it should not apply this label.

rquitales: > Since PKO doesn't know the application to which the workload is associated, it should not apply this label.

I disagree. The objects we are creating here (Workspace, Statefulset) are all created by our own PKO and should also be managed by PKO. These objects are what enables PKO to work right now. What happens if another controller/tool takes over control of these objects? We'd likely get into a controller fight here. Though practically, these labels are most generally used for metrics, so having the `app.kubernetes.io/managed-by` would allow users to find metrics for PKO created/managed objects in their current dashboards.

EronWright: They could also use the pulumi.com labels to identity these pods.

If you were to deploy an EKS cluster with various addons, I would guess that you would not see `app.kubernetes.io/managed-by` on the majority of the objects. It just isn't as common as you're suggesting.

rquitales (approved):
While I disagree with removing `app.kubernetes.io/managed-by`, I won't block given the timeline for v2 release. We can discuss this offline.

PR #837, opened 2025-02-19
Polish for README files

EronWright:
### Proposed changes

- new installation instructions, prioritizing Helm
- new migration guide
- new troubleshooting guide
- installer: renamed `deploy/yaml` to `deploy/quickstart` for clarity
- helm: updated helmdoc tool to 1.14.2
- helm: updated installation instructions (as seen on ArtifactHub) to use the `pulumi-kubernetes-operator` by default
- helm: changed the default log level to `info` (was `error`)
- examples: updated pulumi image
- examples: rename `random-yaml` to `flux-source`
- examples: rename `program` to `program-source`
- examples: rename `workspace` to `custom-workspace`
- examples: new example`git-source`


operator/config/flux/network_policy.yaml
```diff
@@ -17,10 +17,8 @@ spec:
         - namespaceSelector:
             matchLabels:
               kubernetes.io/metadata.name: default
-        - podSelector:
+          podSelector:
```
rquitales, line 20: Is this correct? This **only** allows ingress from workspace pods in the `default` namespace. I believe we still want the `-` in `- podSelector:` since we can have workspace pods in namespaces other than `default`, or whatever namespace value is overridden by `kustomize`.

EronWright: Moved to: https://github.com/pulumi/pulumi-kubernetes-operator/pull/839

I believe this network policy intends to allow traffic from workspace pods in the default namespace only.  At least that's what I wrote in its header a while ago:
> A network policy to allow Pulumi workspaces in the `default` namespace to fetch Flux artifacts from the source-controller in the `flux-system` namespace.

With the dash, I believe the rule would allow traffic from non-workspace pods in the default namespace.

I would agree that opening it up to all namespaces would be convenient but it might be unsafe for the Flux source controller, since there's no other authentication mechanism for artifact URLs.

README.md
```diff
@@ -42,18 +42,14 @@ Deploy the operator to a Kubernetes cluster.
 
 You can use an existing cluster, or [get started](https://www.pulumi.com/docs/get-started/kubernetes/) by creating a new [managed Kubernetes cluster](https://www.pulumi.com/docs/tutorials/kubernetes/#clusters). We will assume that your target Kubernetes cluster is already created and you have configured `kubectl` to point to it.
 
-### Using kubectl
```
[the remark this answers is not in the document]
EronWright: I re-added and expanded on it.

PR #843, opened 2025-02-19
New example: pulumi-ts

EronWright:
Contributes a new example showing how to use Pulumi IAC to manage `Stacks`.

A unique aspect of this example is that it automatically provisions a Pulumi Cloud access token for the stack to use. Note this important limitation of `@pulumi/pulumiservice`:

> While you can use this provider to provision access tokens, you’ll still need to have an access token available to generate an access token with the provider.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+- New example: pulumi-ts [#843](https://github.com/pulumi/pulumi-kubernetes-operator/pull/843)
+
 ## 2.0.0 (2025-02-18)
 
 - Sample network policies [#839](https://github.com/pulumi/pulumi-kubernetes-operator/pull/839)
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+- New example: pulumi-ts [#843](https://github.com/pulumi/pulumi-kubernetes-operator/pull/843)
```
rquitales, line 6: I don't think we need a changelog for this. It isn't adding anything to our release bundle, nor does it affect any features.

examples/pulumi-ts/index.ts
```diff
@@ -0,0 +1,62 @@
...
+        },
+        workspaceTemplate: {
+            spec: {
+                image: "pulumi/pulumi:3.147.0-nonroot",
+            },
+        },
+    },
+}, {dependsOn: [sa, crb, apiSecret]});
+
+// export const stackName = stack.metadata.name;
```
rquitales, line 62: nit: should we uncomment this?

rquitales (approved):

PR #923, opened 2025-05-21
Stabilize the ordering of initContainers

### LINKED ISSUE #774: Use a podTemplate inside the workspaceTemplate
We currently expect a fully-formed Pod spec, but we should use an apply configuration instead.

I have a branch where I started this work but I ran into a number of issues.

https://github.com/pulumi/pulumi-kubernetes-operator/pull/713#discussion_r1794131896

EronWright:
### Proposed changes

This PR is a breaking API change to ensure that a workspace's built-in init containers (bootstrap, fetch) are declared before any extra init containers found in the `podTemplate`. 

Implementation-wise, the desired ordering is achieved with the `$setElementOrder` directive, as documented here:
https://github.com/kubernetes/design-proposals-archive/blob/main/cli/preserve-order-in-strategic-merge-patch.md

### Migration Notes
If your application uses the `podTemplate` to inject init container(s), be aware that they'll run after the `bootstrap` and `fetch` containers. This means that the `/share` volume will be present, and your program code will have been downloaded, before your init container runs.

### Related issues (optional)

Closes #774


operator/internal/controller/auto/utils.go
```diff
@@ -116,3 +119,30 @@ type Event interface {
 func emitEvent(recorder record.EventRecorder, object runtime.Object, event Event, messageFmt string, args ...interface{}) {
 	recorder.Eventf(object, event.EventType(), event.Reason(), messageFmt, args...)
 }
+
+func marshalJSON(v any) (strategicpatch.JSONMap, error) {
+	bytes, err := json.Marshal(v)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal JSON: %w", err)
+	}
+	data := map[string]interface{}{}
```
blampe, line 128: nit
```suggestion
	data := map[string]any{}
```

blampe (approved):

PR #924, opened 2025-05-21
Support for ESC environments in the `Stack` resource

### LINKED ISSUE #814: Support for ESC environments in the `Stack` custom resource
## Hello!

## Issue details

A normal stack config file can contain two top level entries:

```yaml
config:
   <ns>:<key>: <value>
environment:
  - <name-of-esc-environment>
  - <name-of-other-esc-environment>
```

The current custom resource definition for `Stack` contains a property `config` that you can use to configure stack config values. But all usages of the word `environment` in the CRD yaml file refer to environment variables. As far as I can see from the CRD, there is no option to configure the list of ESC enviroments to use for the stack.

### Affected area/feature

PKO `Stack` CRD

EronWright:
### Proposed changes

Adds support for importing ESC environments into your `Stack` object or `Workspace`.

For example, a Stack definition that includes `environment`:

```yaml
apiVersion: pulumi.com/v1
kind: Stack
metadata:
  name: example
  namespace: default
spec:
  stack: example
  environment:
    - default/my-environment
```

Will result in a Pulumi stack configuration file (`Pulumi.example.yaml`) with an `environment` block:

```yaml
environment:
  - default/my-environment
```

Note that not all backends support the use of environments, and the workspace will (repeatedly) fail to initialize in such situations.

### Related issues (optional)

Closes #814


blampe (approved):
Hooray for automation api!

PR #928, opened 2025-05-28
Program status to be fully Flux-compatible

EronWright:
### Proposed changes

Since the Program object is treated as a Flux source by the stack controller, it is important that it have a compatible status block, notably a `.status.artifact.revision`. This PR adds a few fields to the `Artifact` type to align with `source.toolkit.fluxcd.io/v1` (see [fluxcd/source-controller/api/v1/artifact_types.go](https://github.com/fluxcd/source-controller/blob/71e6ca6b0a90ae112f404fc6d864128752dbd078/api/v1/artifact_types.go#L26-L62)).

In particular, the Program status now has a `revision` based on the Program object's `.metadata.generation` field. This is a reasonable value because the program content is derived entirely from the `Program` spec, and because:
```
// Revision is a human-readable identifier traceable in the origin source
// system. It can be a Git commit SHA, Git tag, a Helm chart version, etc.
```

This PR also takes the opportunity to append the revision to the artifact URL, and to verify it in the HTTP handler.  The intention is to prevent serving an unintended generation for a given request (likely causing a digest mismatch).

```yaml
apiVersion: pulumi.com/v1
kind: Program
metadata:
  creationTimestamp: "2025-05-28T23:12:07Z"
  generation: 8
  name: random-yaml
  namespace: default
  resourceVersion: "2569241"
  uid: 36e62b73-4750-4ad0-a441-acb0e79a3257
program:
  outputs:
    password: ${randomPassword.result}
  resources:
    randomPassword:
      properties:
        length: 16
        overrideSpecial: _%@
        special: true
      type: random:RandomPassword
status:
  artifact:
    digest: sha256:d711548592b430d34112a0aa21c45c295c8ac632b1c6a5e760f4b65aa23bcc6a
    lastUpdateTime: "2025-05-29T05:58:33Z"
    path: programs/default/random-yaml/8.tar.gz
    revision: "8"
    size: 227
    url: http://pulumi-kubernetes-operator.pulumi-kubernetes-operator:80/programs/default/random-yaml/8
  observedGeneration: 8
```


EronWright:

- #928 ◀
    - #929

operator/internal/controller/pulumi/program_controller.go
```diff
@@ -143,13 +151,13 @@ func (p *ProgramHandler) HandleProgramServing() http.HandlerFunc {
...
 
-		namespace, name, err := urlPathToProgramIdentifier(programIdentifier)
+		namespace, name, generation, err := urlPathToProgramIdentifier(programIdentifier)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
 
-		proj, err := getProjectFile(context.Background(), p.k8sClient, namespace, name)
+		proj, err := getProjectFile(r.Context(), p.k8sClient, namespace, name, generation)
```
blampe, line 160: Good catch!

blampe (approved):

PR #1012, opened 2025-09-22
Preview mode for Stack API

### LINKED ISSUE #16: Consider adding preview
Consider thinking through if `pulumi preview` is useful and needed in an operator, and how it would work.

e.g. Do we store preview links in CR statuses, or hook up GitHub PR webhooks to provide the [Pulumi GitHub app](https://www.pulumi.com/docs/guides/continuous-delivery/github-actions/#pulumi-github-app) preview-in-a-PR capability, or is this not needed at all for an operator?

EronWright:
### Proposed changes

Proposal is to add a new “preview-only” flag that forces a given Stack object to **run in preview mode only**.  This is NOT update-with-a-preview.

The basic idea is for the user to **create multiple Stack objects for the same underlying Pulumi stack**, with all-but-one operating in a dry-run fashion to illustrate a “what-if” configuration.

Such an arrangement would allow the user to **run previews of a stack against the same or a different branch/tag**, and obtain full status such as the preview link, standard output, and program outputs*.  The user may even vary the stack configuration of a given preview.

Another, simpler way to use this feature would be to “toggle” the preview flag on and off on a given Stack object, to **roll out a change in a tick-tock fashion** with external verification between turns.  The Stack API has strong support for `metadata.generation` to make sense of the status (as each spec change would effect a new generation).

The Stack API remains otherwise unchanged.  In preview mode, the concept of “in-sync” (as reflected by the `Ready` condition) indicates that the preview was successful.

### Testing

1. new e2e test - examples/preview-demo (see [README.md](https://github.com/pulumi/pulumi-kubernetes-operator/pull/1012/files#diff-1800d7c23a9e57cc63b48efc015f10522e636245dc7c1fac495c6900d1ca86f5))
2. new unit test cases for "preview" flag

### Example

```yaml
apiVersion: pulumi.com/v1
kind: Stack
metadata:
 name: random-yaml
spec:
 stack: eron-pulumi-corp/random-yaml
 projectRepo: https://github.com/pulumi/examples
 branch: v1.0
status:
 permalink: https://app.pulumi.com/eron-pulumi-corp/random/random-yaml-git/updates/1
 conditions:
 - lastTransitionTime: "2025-09-18T23:33:04Z"
   message: the stack has been processed and is up to date
   reason: ProcessingCompleted
   status: "True"
   type: Ready
---
apiVersion: pulumi.com/v1
kind: Stack
metadata:
 name: random-yaml-preview-v2.0
spec:
 stack: eron-pulumi-corp/random-yaml
 projectRepo: https://github.com/pulumi/examples
 branch: v2.0
 preview: true
status:
 permalink: https://app.pulumi.com/eron-pulumi-corp/random/random-yaml-git/previews/fc0c65fd-0609-4c57-8170-ac3edbc89d4b
 conditions:
 - lastTransitionTime: "2025-09-18T23:33:04Z"
   message: the stack has been previewed
   reason: ProcessingCompleted
   status: "True"
   type: Ready
```

The series of updates and previews, with updated messages, resembles:

image

### Related issues (optional)

Closes #16


operator/internal/controller/auto/update_controller.go
```diff
@@ -643,26 +644,38 @@ func (s streamReader[T]) Result() (result, error) {
...
 			s.obj.Status.Permalink = link
 		}
-		s.obj.Status.Message = res.GetSummary().Message
 		s.u.progressing.Status = metav1.ConditionFalse
 		s.u.progressing.Reason = UpdateConditionReasonComplete
 		s.u.complete.Status = metav1.ConditionTrue
 		s.u.complete.Reason = UpdateConditionReasonUpdated
-		switch res.GetSummary().Result {
-		case string(apitype.StatusSucceeded):
+
```
EronWright, line 654: Rationale here is that "preview" doesn't return a Summary object, so the overall result is a tad simplified in that case.

examples/preview-demo/README.md
```diff
@@ -0,0 +1,148 @@
+# Preview Mode Example
```
blampe, line 1: This is really good documentation, maybe link to it from the changelog?

scripts/grpcurl
```diff
@@ -0,0 +1,13 @@
...
+export KUBERNETES_TOKEN=$(kubectl create token -n pulumi-kubernetes-operator controller-manager --audience=random-yaml-git.default)
+
+# Check if token generation was successful
+if [ -z "$KUBERNETES_TOKEN" ]; then
+    echo "Error: Failed to generate Kubernetes token"
+    exit 1
+fi
+
+# Add authorization header to grpcurl arguments
+grpcurl -H "Authorization: Bearer $KUBERNETES_TOKEN" "$@"
```
blampe, line 13: What was this for?

blampe (approved):
Just a nit but based on your screenshot it seems we're double-quoting somewhere.

Does `preview: true` interact as expected with destroys as well? Not a blocker if not.

PR #1031, opened 2025-10-17
Fix JSON formatting inconsistency in config item marshaling

EronWright:
## Summary
- Fixed JSON formatting inconsistency when marshaling config items to JSON
- Enhanced test coverage with deeply nested object structures

## Details

The test was failing due to non-deterministic whitespace in JSON output from `protojson.Marshal` due to whitespace randomization. This is a known issue (https://github.com/golang/protobuf/issues/1082).

### Changes
1. **agent/pkg/server/server.go**: Use a three-step approach to ensure deterministic compact JSON.
2. **agent/pkg/server/server_test.go**: Enhanced `TestUnmarshalConfigItemsJson` with deeply nested object structure (4 levels deep) to ensure proper handling of complex configurations

### Testing
- All existing tests pass
- New deeply nested object test case validates the fix works with complex structures containing:
  - Multi-level nested objects (connection.primary.ssl.enabled)
  - Arrays of objects
  - Mixed primitive and complex types

### DIFF
--- a/agent/pkg/server/server_test.go
+++ b/agent/pkg/server/server_test.go
@@ -438,8 +438,24 @@ func TestUnmarshalConfigItemsJson(t *testing.T) {
 	numVal := structpb.NewNumberValue(42)
 	boolVal := structpb.NewBoolValue(true)
 	objVal, _ := structpb.NewValue(map[string]interface{}{
-		"host": "localhost",
-		"port": float64(5432),
+		"connection": map[string]interface{}{
+			"primary": map[string]interface{}{
+				"host": "localhost",
+				"port": float64(5432),
+				"ssl": map[string]interface{}{
+					"enabled": true,
+					"cert":    "/path/to/cert",
+				},
+			},
+			"replica": map[string]interface{}{
+				"host": "replica.localhost",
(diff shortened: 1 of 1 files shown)


agent/pkg/server/server.go
```diff
@@ -425,14 +425,24 @@ func unmarshalConfigItemsJson(project string, items []*pb.ConfigItem) (map[strin
...
+				}
+				// Re-marshal with standard json.Marshal for deterministic compact output
+				jsonBytes, err := json.Marshal(obj)
+				if err != nil {
+					return nil, fmt.Errorf("re-marshaling %q value: %w", item.Key, err)
+				}
 				jsonStr := string(jsonBytes)
 				cv.Value = &jsonStr
-				cv.ObjectValue = val
+				cv.ObjectValue = obj
```
EronWright: I'm actually fixing two bugs here; it was incorrect to use `val` directly as `cv.ObjectValue`, because `cv` is later encoded using `json.Marshal`.  The correct approach is for `cv.ObjectValue` to have a pure Go value, as shown here.

I updated the tests to use a more deeply-nested object, to better verify the protojson goodness.

PR #1107, opened 2026-01-27
Mark workspace as stalled instead of deleting pod on init failure

### LINKED ISSUE #1073: Pulumi workspace endlessly retries when it can't find the backend bucket
### What happened?

A pulumi-workspace pod is in a strange loop where it starts, clones the repository with the pulumi recipe but it breaks with a 404 when attempting to access something related to the state bucket. 

The pod is not "restarting", the server stops, and after a few minutes the controller schedules a new pod repeating the process again. So it ends up cloning the git repo several times per day.

Ideally, there'd be a cleaner failure than this that doesn't wind up costing compute to clone the repo repeatedly.

### Example

N/A

### Output of `pulumi about`

N/A

### Additional context

_No response_

### Contributing

joeduffy:
Edited by @guineveresaenger, Feb 12:

When a k8s Stack Custom Resource (CR) is misconfigured, such that the corresponding Pulumi workspace pod cannot initialize a Pulumi Stack (any failure up to and including `pulumi stack init`), the Workspace pod gets recreated indefinitely, causing the Stack CR to hang forever, and the Workspace pods thrashing and their logs being effectively inaccessible due to the re-creation.

This pull request marks such a Workspace as Stalled - a condition that stops progression inside the Workspace and prevents recreation.
Additionally, changes were made to make sure the Stalled condition and its reason is then passed down to the Stack CR.
A user can now more easily discover what went wrong, change their source with the correct configuration, and re-deploy the Stack CR.
Tests are added for both the Stalled functionality as well as the Stack messaging.

Additionally, I've added to ensure that a Stack that is stuck with a stalled Workspace can be deleted via `kubectl delete` by setting the finalizer in the case where the Workspace failed to initialize. Previously, this was only possible manually via `kubectl patch` which asks too much of a casual user.

I've changed the original commit by Claude to avoid adding more ginkgo tests to the repository, mostly for internal tools consistency.

ORIGINAL MESSAGE BELOW:

### Proposed changes

This PR fixes an issue where workspace initialization failures (such as backend bucket not found) would cause the operator to delete the pod, triggering an endless retry loop as the StatefulSet recreates it.

**Changes:**
- Instead of deleting the pod on installation or stack initialization failures, the workspace is now marked with a `Stalled` condition
- Added logic to clear the `Stalled` condition when re-attempting initialization (e.g., after a spec change that might fix the issue)
- Updated error messages to reflect the new behavior ("marking workspace as stalled")
- Added comprehensive regression tests with a mock gRPC server to verify the new behavior

**Benefits:**
- Prevents endless pod recreation cycles
- Allows users to fix the underlying issue (e.g., create the missing backend bucket) and trigger a new reconciliation
- Provides better visibility into why a workspace is not progressing via the `Stalled` condition

### Related issues

Fixes #1073

https://claude.ai/code/session_01AhJ4REonDGYgravhgwg7HD

### DIFF
--- a/operator/internal/controller/pulumi/stack_controller.go
+++ b/operator/internal/controller/pulumi/stack_controller.go
@@ -942,18 +942,16 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 		// Check if workspace is stalled and propagate to the stack
 		stalledCond := meta.FindStatusCondition(sess.ws.Status.Conditions, autov1alpha1.WorkspaceStalled)
 		if stalledCond != nil && stalledCond.Status == metav1.ConditionTrue {
-			// If the stack is being deleted because the workspace is stalled, we need to
-			// remove its finalizer so delete is possible.
-			// This should not result in any orphaned resources, because all workspace stall
-			// conditions only occur during first-time initialization of the Workspace.
-			if isStackMarkedToBeDeleted {
-				log.Info("Workspace is stalled during deletion; skipping destroy and removing finalizer",
+			// Allow for stack deletion if no updates were made successfully.
+			if isStackMarkedToBeDeleted && instance.Status.LastUpdate == nil {
+				log.Info("Workspace is stalled during deletion with no prior successful update; removing finalizer",
 					"reason", stalledCond.Reason, "message", stalledCond.Message)
 				if controllerutil.RemoveFinalizer(instance, PulumiFinalizer) {
 					return reconcile.Result{}, r.Update(ctx, instance, client.FieldOwner(FieldManager))
 				}
 				return reconcile.Result{}, nil
 			}
+
 			var reason string
 			if stalledCond.Reason == "IncompatibleConfiguration" {
 				reason = pulumiv1.StalledPulumiVersionTooLowReason


guineveresaenger:
hi @joeduffy - are you able to take a look at the failing test here?

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -939,12 +939,30 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
+			// This should not result in any orphaned resources, because all workspace stall
+			// conditions only occur during first-time initialization of the Workspace.
+			if isStackMarkedToBeDeleted {
+				log.Info("Workspace is stalled during deletion; skipping destroy and removing finalizer",
+					"reason", stalledCond.Reason, "message", stalledCond.Message)
+				if controllerutil.RemoveFinalizer(instance, PulumiFinalizer) {
+					return reconcile.Result{}, r.Update(ctx, instance, client.FieldOwner(FieldManager))
+				}
+				return reconcile.Result{}, nil
+			}
```
blampe: > // If the stack is being deleted because the workspace is stalled, we need to
> // remove its finalizer so delete is possible.

What if the workspace if stalled be because it's unable to destroy the stack, say due to a protected resource or something?

blampe (commented):
https://github.com/pulumi/pulumi-kubernetes-operator/issues/801 might have some historical context. In that case the stall was due to an invalid image. #802 came with an e2e test, and if that's still passing after this change it's a good sign we haven't regressed anything.

There's always been some tension between these two failure modes -- "thing gets periodically recreated" vs. "thing sits around doing nothing". My understanding is we're switching this from the former to the latter, but some users _really_ don't like the latter because they're sensitive to wasted resource reservations.

There is `Stack.workspaceReclaimPolicy` to help make everyone happy. It defaults to Retain but also allows Delete. You might consider respecting here so users can get whichever behavior they prefer.

> I've changed the original commit by Claude to avoid adding more ginkgo tests to the repository, mostly for internal tools consistency.

Thank you 🙌

PR #1251, opened 2026-06-24
Surface a Stalled condition when a Stack can't be destroyed in a terminating namespace

guineveresaenger:
This pull request adds a troubleshooting messagewhen a Stack with DestroyOnFinalize is hung due to being in a terminating Namespace.

Workspace pods need to be newly created for each pulumi operation, but when a Kubernetes Namespace is in Terminating state, no new object creations are possible.
Best practice says to manually order a Namespace object to be deleted _after_ its CRDs, but we can help out the user by logging a warning and troubleshooting steps.

Addresses #1181.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/operator/internal/controller/pulumi/stack_controller_test.go
+++ b/operator/internal/controller/pulumi/stack_controller_test.go
@@ -3056,6 +3056,107 @@ func TestFailedDestroyRetainsFinalizer(t *testing.T) {
 	assert.Contains(t, updated.Finalizers, PulumiFinalizer)
 }
 
+func TestStackDestroyStalledWhenNamespaceTerminating(t *testing.T) {
+	testScheme := scheme.Scheme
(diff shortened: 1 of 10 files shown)


operator/internal/controller/pulumi/stack_controller_test.go
```diff
@@ -2992,6 +2992,107 @@ func TestProjectInfoDestroyBypassesUnavailableSource(t *testing.T) {
 	assert.Equal(t, autov1alpha1.DestroyType, update.Spec.Type, "Update CR must be of type destroy")
 }
 
+func TestStackDestroyStalledWhenNamespaceTerminating(t *testing.T) {
```
pose, line 3059: Wow, that's a lot of things to orchestrate to get to repro!

pose (approved):

PR #1252, opened 2026-06-24
Re-generate CRDs and add CI gate

guineveresaenger:
The published resources for recent API additions were drifting.

This pull request re-generates them properly, and adds a CI check to prevent drift in the future.

- **Re-generate CRDs for new ProjectInfo field**
- **Add CI gate for codegen diffs**


nyobe (approved):
:stamp:

PR #1253, opened 2026-06-24
Retain finalizer and retry when destroyOnFinalize destroy fails

### LINKED ISSUE #1233: destroyOnFinalize removes Stack finalizer after failed destroy during retry cooldown
## What happened?

A `Stack` CR with `spec.destroyOnFinalize: true` can be finalized and removed after the destroy `Update` fails.

This is related to, but distinct from #1220 / #1223. #1223 fixes the case where a successful prior `up` is treated as "already destroyed" and no destroy `Update` is created. In this case, the destroy `Update` is created and fails, but the Stack controller still removes `finalizer.stack.pulumi.com`.

## Reproduction

1. Run PKO v2.3.0 with a Stack CR using:

```yaml
spec:
  destroyOnFinalize: true
```

2. Let the Stack successfully reconcile with `pulumi up`.
3. Delete the Stack CR:

```bash
kubectl delete stack <name> --wait=false
```

4. Ensure the destroy operation fails. In our repro, a Pulumi dynamic resource delete failed:

```text
pulumi-python:dynamic:Resource (cluster-teardown):
  error: Exception calling application: HTTP Error 422: Unprocessable Entity
```

5. Observe the Stack CR and downstream resources.

## What actually happened?

PKO created a destroy `Update`, the destroy failed, and the Stack status showed:

```text
lastUpdate.type: destroy
lastUpdate.state: failed
```

PKO then logged:

```text
Stack was destroyed; finalizing now.
```

and removed `finalizer.stack.pulumi.com`.

The Stack CR disappeared, the failed `Update` disappeared, and downstream resources remained orphaned.

In our case, CAPI/CAPD resources were still present after the Stack CR vanished:

```text
Cluster/example-workload
Machine/example-workload-*
DockerMachine/example-workload-*
DockerCluster/example-workload-*
```

## What should happen?

If `destroyOnFinalize` destroy fails, PKO should keep the Stack CR terminating with the finalizer intact, surface the failed destroy status/events, and retry according to backoff.

The finalizer should only be removed when the last update is both:

```text
lastUpdate.type == destroy
lastUpdate.state == succeeded
```

## Suspected root cause

After #1223, the successful-update deletion shortcut is guarded by `LastUpdate.Type == DestroyType`, which fixes the no-destroy-created path.

However, the failed-update branch in `isSynced()` still treats failed updates as synced while inside retry cooldown:

```go
if stack.Status.LastUpdate.State == shared.FailedStackStateMessage {
    ...
    c := cooldown(stack)
    if time.Since(stack.Status.LastUpdate.LastResyncTime.Time) >= c {
        return false, msg
    }
    return true, ""
}
```

The caller then sees:

```go
synced == true && isStackMarkedToBeDeleted
```

and removes the finalizer.

Code inspection of upstream `master` after #1223 suggests this failed-destroy cooldown path is still reachable because the failed-update branch still returns `true` during cooldown.

## Version

Observed live on PKO v2.3.0.

## Logs

Relevant operator log excerpt:

```text
Executing destroy operation {"type":"destroy"}
...
failed to run destroy
...
Status updated ... "type":"destroy","state":"failed"
...
Stack was destroyed; finalizing now.
```

Destroy failure excerpt:

```text
pulumi-python:dynamic:Resource cluster-teardown deleting
error: Exception calling application: HTTP Error 422: Unprocessable Entity
pulumi:pulumi:Stack ... error: update failed
Resources:
  - 4 deleted
  2 errored
```

## Why this matters

This silently orphans resources on failed destroy. For infrastructure stacks, a failed finalizer should be loud and blocking, not remove the CR and erase the retry surface.

guineveresaenger:
When a `destroyOnFinalize: true` Stack's `pulumi destroy` fails, the operator currently removes the Pulumi finalizer on the same reconcile that observes the failure — deleting the Stack and silently orphaning its cloud resources after a single failed attempt (the log even claims "Stack was destroyed"). This happens because `isSynced` returns true during the post-failure cooldown, and the finalize branch removed the finalizer without checking whether the destroy actually succeeded.

This guards finalizer removal on `LastUpdate.State == Succeeded`. A failed destroy now retains the finalizer and retries with the existing exponential backoff, exactly like a failed update — surfaced via `Reconciling / RetryingAfterFailure` and the Update's error, never silently abandoned. The deliberate escape hatch remains a manual `kubectl patch stack … finalizers:[]`.

Closes #1233.

## Test plan

- `TestFailedDestroyRetainsFinalizer` (envtest): seeds a failed destroy in its backoff cooldown, reconciles, and asserts the Pulumi finalizer is retained. Verified it **fails against `origin/master`** (finalizer stripped → orphan) and **passes with this change**.
- Full `internal/controller/pulumi` package green via `KUBEBUILDER_ASSETS`.

### DIFF
--- a/operator/internal/controller/pulumi/stack_controller.go
+++ b/operator/internal/controller/pulumi/stack_controller.go
@@ -812,6 +812,15 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 		}
 
 		if isStackMarkedToBeDeleted {
+			// Only finalize if the destroy succeeded; a failed destroy must retain the
+			// finalizer and retry, not silently orphan the stack's resources.
+			if instance.Status.LastUpdate.State != shared.SucceededStackStateMessage {
+				log.Info("Destroy failed; retaining finalizer and retrying after backoff",
+					"failures", instance.Status.LastUpdate.Failures, "message", instance.Status.LastUpdate.Message)
+				requeueAfter := max(1*time.Second, time.Until(instance.Status.LastUpdate.LastResyncTime.Add(cooldown(instance))))
+				return reconcile.Result{RequeueAfter: requeueAfter}, saveStatus()
+			}
+
 			log.Info("Stack was destroyed; finalizing now.")
 			_ = saveStatus()
 			if controllerutil.RemoveFinalizer(instance, PulumiFinalizer) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 ## Unreleased
 
 - Fix `destroyOnFinalize: true` silently skipping `pulumi destroy` for any Stack that had been successfully reconciled [#1223](https://github.com/pulumi/pulumi-kubernetes-operator/pull/1223)
+- Fix `destroyOnFinalize: true` removing the finalizer and silently orphaning a Stack's cloud resources when `pulumi destroy` fails; the operator now retains the finalizer and retries the destroy with backoff instead of deleting the Stack [#1233](https://github.com/pulumi/pulumi-kubernetes-operator/issues/1233)
 - Fix `destroyOnFinalize: true` getting stuck in `Stalled / SourceUnavailable` when a Stack is deleted together with its source (an inline `Program` or a Flux source); the operator now destroys from backend state without re-fetching the source [#1222](https://github.com/pulumi/pulumi-kubernetes-operator/issues/1222) [#441](https://github.com/pulumi/pulumi-kubernetes-operator/issues/441)
 
 ## 2.7.0 (2026-04-06)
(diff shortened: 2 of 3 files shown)


operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -812,6 +812,15 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 		}
 
 		if isStackMarkedToBeDeleted {
+			// Only finalize if the destroy succeeded; a failed destroy must retain the
+			// finalizer and retry, not silently orphan the stack's resources.
+			if instance.Status.LastUpdate.State != shared.SucceededStackStateMessage {
+				log.Info("Destroy failed; retaining finalizer and retrying after backoff",
```
pose, line 818: (minor) maybe this should be a `warn` instead?

pose (approved):

PR #1255, opened 2026-06-25
helm: make controller liveness/readiness probes configurable

### LINKED ISSUE #1254: Controller liveness/readiness probe timeouts are not configurable (hardcoded to Kubernetes default of 1s)
## What happened?

The controller-manager `Deployment` in the Helm chart hardcodes its liveness and readiness probes and never sets `timeoutSeconds` or `failureThreshold`:

https://github.com/pulumi/pulumi-kubernetes-operator/blob/master/deploy/helm/pulumi-operator/templates/deployment.yaml#L98-L109

Because `timeoutSeconds` is unset, it falls back to the Kubernetes default of **`timeoutSeconds: 1`** (with `failureThreshold: 3`).

When the manager is busy during long-running Pulumi operations, the `/healthz` endpoint can fail to respond within that 1-second window. After 3 consecutive misses (~`3 × periodSeconds`), the kubelet sends SIGTERM (the container exits with code 143) and the **in-flight update is cancelled mid-run**. On restart the controller can re-enter the same condition, and the cancellation can leave behind orphaned state locks / stuck workspaces that the controller does not self-recover from.

Both `/healthz` and `/readyz` are wired to `healthz.Ping`:

https://github.com/pulumi/pulumi-kubernetes-operator/blob/master/operator/cmd/main.go#L307-L312

so the probe handler itself is trivial — the failures are caused by the process being momentarily starved, which a 1s timeout does not tolerate.

## Context

PR #1059 added configurable leader-election timeouts for exactly this "don't interrupt long-running operations" reason, but the probe timeouts were left at the aggressive Kubernetes defaults. Raising the pod's CPU/memory limits does not help, since the kill is driven by probe latency (SIGTERM/143), not OOM.

## Expected

Operators should be able to tune the liveness/readiness probes (`timeoutSeconds`, `failureThreshold`, `periodSeconds`, `initialDelaySeconds`) via `values.yaml`, the same way leader-election timeouts are already configurable — without forking the chart.

## Proposed fix

Make the probes configurable in the chart, preserving the current defaults (`timeoutSeconds: 1`, `failureThreshold: 3`) so there is no behavior change for existing users. PR: see below.

## Environment

- Chart: `deploy/helm/pulumi-operator`
- Operator: v2.7.0

aashishshrestha09:
## Summary

Fixes #1254.

The controller-manager `Deployment` hardcodes its liveness and readiness probes and never sets `timeoutSeconds`/`failureThreshold`, so they fall back to the Kubernetes default of `timeoutSeconds: 1`. When the manager is busy during long-running Pulumi operations, `/healthz` can fail to answer within 1s; after the default 3 consecutive failures the kubelet SIGTERMs the container (exit 143) and cancels the in-flight update mid-run.

This mirrors the long-running-operation concern already addressed for leader election in #1059, but the probe timeouts were left at the aggressive Kubernetes defaults.

## Changes

- Expose `livenessProbe` and `readinessProbe` (`initialDelaySeconds`, `periodSeconds`, `timeoutSeconds`, `failureThreshold`) in `values.yaml`.
- Wire them into `templates/deployment.yaml`.

## Backwards compatibility

Defaults are preserved (`timeoutSeconds: 1`, `failureThreshold: 3`), so there is **no behavior change** for existing users. Operators running long Pulumi operations can opt into more tolerant values, e.g.:

```yaml
livenessProbe:
  timeoutSeconds: 5
  failureThreshold: 5
readinessProbe:
  timeoutSeconds: 5
  failureThreshold: 5
```

## Testing

`helm template` renders the current defaults unchanged, and honors overrides:

- defaults → `timeoutSeconds: 1`, `failureThreshold: 3`
- `--set livenessProbe.timeoutSeconds=5 --set livenessProbe.failureThreshold=5` → `timeoutSeconds: 5`, `failureThreshold: 5`

### DIFF
--- a/deploy/helm/pulumi-operator/templates/deployment.yaml
+++ b/deploy/helm/pulumi-operator/templates/deployment.yaml
@@ -99,14 +99,18 @@ spec:
           httpGet:
             path: /healthz
             port: 8081
-          initialDelaySeconds: 15
-          periodSeconds: 20
+          initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
+          timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
+          failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
         readinessProbe:
           httpGet:
             path: /readyz
             port: 8081
-          initialDelaySeconds: 5
-          periodSeconds: 10
+          initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.readinessProbe.periodSeconds }}
+          timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds }}
+          failureThreshold: {{ .Values.readinessProbe.failureThreshold }}
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
--- a/deploy/helm/pulumi-operator/values.yaml
+++ b/deploy/helm/pulumi-operator/values.yaml
@@ -50,6 +50,26 @@ kubeAPI:
   # Increase this value if experiencing kube-apiserver slowness that causes leader election failures
   timeout: "30s"
 
+# Liveness probe for the controller manager (/healthz on :8081)
+# Increase timeoutSeconds/failureThreshold if long-running Pulumi operations can briefly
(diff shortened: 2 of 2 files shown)


piercenat (approved):

guineveresaenger (commented):
Hi, thank you for this pull request. Because #1254 is an enhancement request, not a bug, we want to make sure that this is implemented correctly.

One thing to consider: should the full probe be configurable?

guineveresaenger:
hi @aashishshrestha09 - are you still interested in working on this? if so, let's discuss details in #1254!

Issue #1007, opened 2025-09-16
ArgoCD Health Checks


EronWright:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Develop and test some Lua-based health check scripts for ArgoCD, to be offered in two ways:
1. document how to add a health check to the ArgoCD server's configmap
2. contribute upstream to the official health checks

The goal is to teach ArgoCD to understand the progression of the deployment, and the health of the application based on whether drift is detected, etc.

guineveresaenger:
https://github.com/pulumi/pulumi-kubernetes-operator/pull/1259 addresses the functionality/
I've opened http://github.com/argoproj/argo-cd/pull/28534 to hopefully add health checks to Argo outright.
Internal docs for how to set it up manually: https://github.com/pulumi/pulumi-kubernetes-operator/pull/1263

This issue can be closed once https://github.com/pulumi/pulumi-kubernetes-operator/pull/1263 and http://github.com/argoproj/argo-cd/pull/28534 are merged.

guineveresaenger:
Update: https://github.com/pulumi/pulumi-kubernetes-operator/pull/1263 was merged. Awaiting https://github.com/argoproj/argo-cd/pull/28534 which depends on Argo maintainer availability so labeling this as awaiting upstream.

PR #1259, opened 2026-06-26
Introduce Maximum Retries and related failure modes

guineveresaenger:
Currently, when a `pulumi up` fails inside the Workspace, the Operator continues to try to reconcile the operation, continuously kicking off new Update CRs at larger and larger backoff periods. The Stack Status reads `Reconciling=True/RetryingAfterFailure`, forever. 

But most of the time, when a pulumi up fails, the failure is deterministic and repeated. A new `pulumi up` will most of the time not somehow succeed where it failed before.

When you hook up the Operator to ArgoCD or any kind of health check, the health check will show a Progressing stack, which is misleading in the case of repeated `pulumi up` failures.

This pull request adds a hardcoded (for now) maximum update failure const of 3. If `pulumi up` fails to succeed 3 times in a row, the Stack Status flips from Progressing to Stalled. Integrations will now show a closer version of the truth. To aid with surfacing the correct failure reasons, `ReconcilingPreviewFailedReason` and `StalledUpdateFailedReason` were added.

This work satisfies the code/logic parts of https://github.com/pulumi/pulumi-kubernetes-operator/issues/1007. With this logic in place, a standard Lua health check will report the reality of "your pulumi up failed" pretty consistently on ArgoCD. Other tooling should benefit as well (although I've only tried this with Argo + Lua).

### DIFF
--- a/operator/internal/controller/pulumi/stack_controller.go
+++ b/operator/internal/controller/pulumi/stack_controller.go
@@ -805,11 +805,7 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 	if synced {
 		// We don't mark the stack as ready if its update failed so downstream
 		// Stack dependencies aren't triggered.
-		if instance.Status.LastUpdate.State == shared.SucceededStackStateMessage {
-			instance.Status.MarkReadyCondition()
-		} else {
-			instance.Status.MarkReconcilingCondition(pulumiv1.ReconcilingRetryReason, fmt.Sprintf("%d update failure(s)", instance.Status.LastUpdate.Failures))
-		}
+		markStackResult(&instance.Status)
 
 		if isStackMarkedToBeDeleted {
 			log.Info("Stack was destroyed; finalizing now.")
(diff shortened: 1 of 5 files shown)


operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -1049,6 +1048,39 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 	return reconcile.Result{}, nil
 }
 
+// maxUpdateFailures bounds how many times a failing update is retried before the
+// operator gives up. Below the limit a failure is treated as transient: the Stack stays
+// Reconciling and retries with backoff.
+//
+// TODO: promote to a spec field.
+const maxUpdateFailures = 3
```
nyobe, line 1056: 👍  K8s Jobs use a [`backoffLimit`](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy) with a default of 6

nyobe (approved):

operator/internal/controller/pulumi/stack_controller.go
```diff
@@ -1049,6 +1048,39 @@ func (r *StackReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 	return reconcile.Result{}, nil
 }
 
+// maxUpdateFailures bounds how many times a failing update is retried before the
+// operator gives up. Below the limit a failure is treated as transient: the Stack stays
+// Reconciling and retries with backoff.
+//
+// TODO: promote to a spec field.
```
dmassecoveo, line 1055: @guineveresaenger FYI, I have opened a PR to add a spec field for maxUpdateFailures as we had use cases where the 3 retry was not enough.

https://github.com/pulumi/pulumi-kubernetes-operator/pull/1322

PR #1280, opened 2026-07-23
Add serviceTemplate to Workspace spec for Service annotations/labels

bpalermo:
## Description

Adds a `serviceTemplate` field to the Workspace spec so users can set custom **annotations and labels** on the headless `Service` that fronts a workspace's pods.

Previously there was no path to annotate that Service: it is fully synthesized by `newService` in the Workspace controller (only system labels, no annotations), it inherits nothing from the owning Workspace's metadata, and `stack.spec.workspaceTemplate` could not reach it because the Service is not part of the Workspace API surface.

Because `StackSpec.WorkspaceTemplate` is a Workspace apply configuration, the new field is settable from a Stack with no new Stack-side plumbing:

```yaml
apiVersion: pulumi.com/v1
kind: Stack
spec:
  workspaceTemplate:
    spec:
      serviceTemplate:
        metadata:
          annotations:
            prometheus.io/scrape: "true"
          labels:
            team: platform
```

## Implementation notes

- New `ServiceTemplate *EmbeddedServiceTemplateSpec` field on `WorkspaceSpec`, reusing the existing `EmbeddedObjectMeta` (annotations + labels), mirroring `PodTemplate`.
- `newService` merges the template labels/annotations onto the Service metadata. **System labels win over conflicting template labels**, and the `Spec.Selector` is kept to the system labels only, so pod routing is never affected.
- Regenerated deepcopy, apply configurations, CRDs (Workspace + Stack), and API docs.

## Testing

- Added `TestNewService` covering the no-template case, the merge case, and selector isolation.
- `go vet` clean.


operator/internal/controller/auto/workspace_controller.go
```diff
@@ -921,6 +933,22 @@ func newService(w *autov1alpha1.Workspace) *corev1.Service {
 	return service
 }
 
+// mergeStringMaps returns a new map containing all entries from base, with
+// entries from override applied on top. It returns nil if both are empty.
+func mergeStringMaps(base, override map[string]string) map[string]string {
```
guineveresaenger: Any reason not to use https://github.com/kubernetes/apimachinery/blob/v0.34.1/pkg/labels/labels.go#L127?

operator/api/auto/v1alpha1/workspace_types.go
```diff
@@ -284,6 +291,14 @@ type EmbeddedPodTemplateSpec struct {
 	Spec *corev1.PodSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
 }
 
+// EmbeddedServiceTemplateSpec is a reduced template for the Workspace's Service.
+// It contains only the metadata that is safe to customize.
+type EmbeddedServiceTemplateSpec struct {
+	// EmbeddedMetadata contains metadata relevant to an embedded resource.
```
guineveresaenger: can we match the object name in the comment?

operator/internal/controller/auto/workspace_controller_test.go
```diff
@@ -1028,6 +1028,64 @@ func TestNewStatefulSet_ProjectInfoSource(t *testing.T) {
...
+				Annotations: map[string]string{"prometheus.io/scrape": "true"},
+				Labels:      map[string]string{"team": "platform"},
+			},
+		}
+
+		svc := newService(wt)
+
+		assert.Equal(t, "true", svc.Annotations["prometheus.io/scrape"])
+		assert.Equal(t, "platform", svc.Labels["team"])
+		// System labels are still present on the metadata.
```
guineveresaenger: superfluous as per assertion below

guineveresaenger (changes requested):
Hi @bpalermo, thank you for this PR.
I was curious about what is your use case? We don't have an existing issue tracking this, and it'd help us understand the need before we commit to a new API field. Thank you so much!

bpalermo:
> Hi @bpalermo, thank you for this PR.
> I was curious about what is your use case? We don't have an existing issue tracking this, and it'd help us understand the need before we commit to a new API field. Thank you so much!
> 
> 

@guineveresaenger currently the headless services get picked by Istio, and causes a lot of churn to our control plane due to the number of stacks in the cluster.
We want to be able to annotate them in order to exclude:
```
networking.istio.io/exportTo: "~"
```

operator/internal/controller/auto/workspace_controller.go
bpalermo: Good call — switched to `labels.Merge` and dropped the custom `mergeStringMaps` helper. Renamed the local `labels` var to `systemLabels` to avoid shadowing the new import.

operator/api/auto/v1alpha1/workspace_types.go
bpalermo: Done — the comment now leads with `Metadata` to match the field, and I regenerated the Workspace CRD/docs accordingly.

operator/internal/controller/auto/workspace_controller_test.go
bpalermo: Removed — dropped the per-label loop (and the redundant selector `NotContains`) in favor of a single full-map equality assertion.

guineveresaenger:
@bpalermo - ah, that makes sense.

LGTM - tests are running

guineveresaenger (approved):

Issue #1297, opened 2026-08-07
pulumi install runs on every update even when the program is a prebuilt binary


bpalermo:
**Version:** v2.9.0

## Summary

The workspace controller calls the agent's `Install` RPC unconditionally on every Workspace reconcile, with no check for whether the project actually needs dependency resolution. For a project using `runtime.options.binary` — a prebuilt program — there is nothing to install, yet a failure there marks the Workspace `InstallationFailed` and **no update ever runs**.

The `--skip-install` flag does not help: it gates only the agent's serve-time path, not the RPC.

```go
// operator/internal/controller/auto/workspace_controller.go:324
l.Info("Running pulumi install")
ready.Reason = "Installing"
_, err = wc.Install(ctx, &agentpb.InstallRequest{})
if err != nil {
    l.Error(err, "unable to install; marking workspace as stalled")
    ready.Reason = "InstallationFailed"
```

```go
// agent/cmd/serve.go:154 — the flag the operator passes at :533
if !_skipInstall {
    ... auto.InstallOptions ...
}
```

Both lines appear in a single workspace's log, which is what makes the behaviour confusing in practice:

```
INFO  cmd.serve  installation skipped   {"project":"...","runtime":"go"}   ← startup, --skip-install honoured
INFO  server     installing the project dependencies                        ← per reconcile, via the RPC
ERROR server     install completed with an error
```

## Why it matters beyond wasted work

We moved a Go project to a prebuilt binary (`runtime.options.binary`) specifically to stop compiling inside the workspace — the compile peaked near 1 GiB RSS, was OOM-killed, and died holding the S3 backend lock, wedging most of a ~290-Stack fleet.

Because `install` still runs, the workspace **still needs a complete Go environment**, and each missing piece is a total outage rather than a degradation. Three separate failures, each surfacing only after the previous was fixed:

| | failure | why |
|---|---|---|
| 1 | `determining go version: exit status 1` | `go.mod` required a newer toolchain than the image shipped, so Go tried to download one; the workspace has no egress to the module proxy |
| 2 | `failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied` | the operator forces `runAsUser: 1000`; the image had no home for that uid, so `HOME=/` and `GOCACHE=/.cache/go-build` |
[...]

PR #1306, opened 2026-08-14
feat(agent): opt-in JSON logging via AGENT_JSON_LOG env var

### LINKED ISSUE #952: Allow configure workspace to run pulumi with --json
## Hello!

## Issue details

In the pulumi-operator v1, logs weren't great but at least they were structured. In V2 we cant configure the `--zap` anymore to change the logging format and is using "console" what is terrible for parsing and sending to log backend (e.g. opensearch).

The pulumi operator itself does have the option to chose `json` in the helm char values. The workspace should have the same option and not just the log level.

In addition, it would be good to be able to change everything via ENV_VAR if there is no option in the stack (I think this is not a stack option but a more generic "always use structured logging"

### Affected area/feature

Pulumi kubernetes operator V2

MsfPablo:
Closes #952.

### What & why

The `agent` binary that runs in every `Workspace` pod (the gRPC executor for stack operations) hard-codes `zap.NewDevelopmentConfig()`, so log output is always the human-readable console encoder. Operators running log-shipping pipelines (Loki/Elastic/Datadog/etc.) currently have to run a separate parser or patch the binary.

This PR adds an opt-in switch: setting `AGENT_JSON_LOG=true` on the pod flips the encoder to JSON. Anything other than the literal string `"true"` defaults to the historical console encoder — preserving existing workflows and log scrapers.

The value can be threaded in any of three existing pathways:

| Path | How |
|------|-----|
| Per-Workspace (user-facing) | `spec.env[].name = AGENT_JSON_LOG`, `value: "true"` |
| Helm chart | `extraEnv: [{name: AGENT_JSON_LOG, value: "true"}]` |
| Direct pod manifest | `env: [{name: AGENT_JSON_LOG, value: "true"}]` |

No controller, CRD, or chart-API changes. The operator just learns to read its env.

### Implementation

- `agent/cmd/logging.go`: new helper `applyJSONLogMode(zc zap.Config) zap.Config` — pure, no side effects, pass-by-value so callers retain full control.
- `agent/cmd/root.go`: in `PersistentPreRunE`, after the level check, call `zc = applyJSONLogMode(zc)`.
- `agent/cmd/logging_test.go`: table-driven tests covering unset, off-values (`false`, `0`, `no`, `TRUE ` with trailing space, `yes`, `True`), and the `true` gating case.

### Test plan

- [x] `go test ./agent/cmd/ -run TestApplyJSONLogMode -v` → all 8 sub-tests pass.
- [x] `go build ./...` → full repo builds clean.
- [ ] (maintainer) run the agent in a sample workspace with and without `AGENT_JSON_LOG=true`; confirm output format change.

### Maintenance note

This is the minimum-cut change. A follow-up could expose it as a proper CRD field on `WorkspaceSpec` (e.g. `jsonLogging: true`) for validation and ergonomics, but that would require a CRD bump. Keeping this PR scoped to the agent side so it lands quickly and unblocks the existing-scrapers use case today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## Unreleased
 
+- Skip `pulumi install` for a project that runs a prebuilt binary (`runtime.options.binary`) [#1297](https://github.com/pulumi/pulumi-kubernetes-operator/issues/1297)
+- Add opt-in JSON log output for the workspace agent, enabled by setting `AGENT_JSON_LOG=true` [#952](https://github.com/pulumi/pulumi-kubernetes-operator/issues/952)
+
 ## 2.9.0 (2026-07-29)
 
 - Add `serviceTemplate` to the Workspace spec, allowing custom annotations and labels on the headless Service that fronts a workspace's pods; settable from a Stack via `spec.workspaceTemplate.spec.serviceTemplate.metadata` [#1280](https://github.com/pulumi/pulumi-kubernetes-operator/pull/1280)
--- a/agent/cmd/logging.go
+++ b/agent/cmd/logging.go
@@ -1,0 +1,36 @@
+// Copyright 2016-2025, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


MsfPablo:
Added the CHANGELOG entry for the `Update Changelog` check.

Heads up: it now shows a conflict in CHANGELOG.md, and it's only that my line lands immediately under the `#1297` entry that went into master after this branch was cut — the two additions touch the same spot in an otherwise empty `## Unreleased` section. I can't merge master in from my side to settle it (my token can't push the workflow file changes that come with it), so it may be easiest to resolve in the web editor or just squash-merge over it.

PR #1319, opened 2026-09-01
Skip pulumi install for projects that run a prebuilt binary

### LINKED ISSUE #1297: pulumi install runs on every update even when the program is a prebuilt binary

guineveresaenger:
Fixes #1297.

Adds detection of `runtime.options.binary` being set and skips calling the install RPC accordingly.

Includes tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ CHANGELOG
 
 ## Unreleased
 
+- Skip `pulumi install` for a project that runs a prebuilt binary (`runtime.options.binary`) [#1297](https://github.com/pulumi/pulumi-kubernetes-operator/issues/1297)
+
 ## 2.9.0 (2026-07-29)
 
 - Add `serviceTemplate` to the Workspace spec, allowing custom annotations and labels on the headless Service that fronts a workspace's pods; settable from a Stack via `spec.workspaceTemplate.spec.serviceTemplate.metadata` [#1280](https://github.com/pulumi/pulumi-kubernetes-operator/pull/1280)
--- a/agent/cmd/serve.go
+++ b/agent/cmd/serve.go
@@ -151,7 +151,7 @@ var serveCmd = &cobra.Command{
 		log.Infow("opened a local workspace", "workspace", workDir,
 			"project", proj.Name, "runtime", proj.Runtime.Name())
 
-		if !_skipInstall {
+		if !_skipInstall && !server.HasPrebuiltBinary(proj) {
 			plog := zap.L().Named("pulumi")
 			stdout := &zapio.Writer{Log: plog, Level: zap.InfoLevel}
 			defer stdout.Close()
--- a/agent/pkg/server/server.go
+++ b/agent/pkg/server/server.go
@@ -48,6 +48,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/config"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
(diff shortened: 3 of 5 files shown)


guineveresaenger:
@pulumi-eon please review

corymhall (approved):

PR #1320, opened 2026-09-01
Don't run pulumi install on project-info workspaces

### LINKED ISSUE #1299: pulumi install runs on project-info workspaces and blocks source-free destroy
**Version:** v2.9.0

## Summary

The source-free destroy path introduced in #1222 creates a Workspace with `spec.projectInfo`, which bootstraps from a synthesized `Pulumi.yaml` containing only `name` and `runtime` — deliberately no program. The workspace controller then runs `pulumi install` on it anyway, because the Install RPC is called unconditionally.

There is nothing to install in that directory, and for every runtime except `yaml` the attempt fails. The Workspace goes `Stalled/InstallationFailed`, the Stack goes `Stalled/WorkspaceFailed`, **the finalizer is retained and the destroy never runs** — which is the resource-orphaning outcome #1222 and #1233 were meant to prevent.

Found while reviewing #1297; filing separately because it is a distinct failure with a distinct trigger.

## Why the usual escape hatch does not apply

`stack_controller.go` releases the finalizer when a workspace stalls during deletion, but only if the Stack never succeeded:

```go
// stack_controller.go:997
if isStackMarkedToBeDeleted && instance.Status.LastUpdate == nil {
    ... RemoveFinalizer ...
}
```

Reaching the project-info path requires `stack.Status.ProjectInfo != nil` (`shouldUseProjectInfoWorkspaceForDestroy`), and that field is only ever set by `markStackSucceeded` (`stack_controller.go:1178`). So `LastUpdate` is necessarily non-nil, the branch is skipped, and the Stack lands on `MarkStalledCondition(StalledWorkspaceFailedReason)` with its finalizer intact.

## Errors & Logs

The project-info workspace writes exactly this (`agent/cmd/init.go:158-170`, note `NewProjectRuntimeInfo(runtime, nil)` — no options):

```yaml
name: <project>
runtime: go
```

That file is byte-identical in shape to the repo's own `agent/pkg/server/testdata/uninstallable` fixture, whose test asserts install fails. Confirmed directly against Pulumi v3.256.0, one temp dir per runtime containing only the above:

| runtime | `pulumi install` |
|---|---|
| `go` | ``error: installing dependencies: `go mod tidy` failed to install dependencies: exit status 1`` (`go.mod file not found`) |
| `nodejs` | `exit status 254` |
| `dotnet` | `unable to find program: dotnet` |
| `yaml` | succeeds |

So the bug is invisible for YAML programs and hits everything else.

## Reproducing the issue

1. Create a Stack with `destroyOnFinalize: true` and a non-YAML runtime (e.g. `go`), and let it complete one successful update, so `status.projectInfo` is populated.
2. Remove its source — delete the inline `Program`, or make the Flux source unavailable.
3. Delete the Stack.
4. The operator takes the project-info bypass, the Workspace stalls at `InstallationFailed`, and the Stack sits at `Stalled/WorkspaceFailed` with `finalizer.stack.pulumi.com` retained. The cloud resources are never destroyed.

## Expected

The operator should not run `pulumi install` for a workspace bootstrapped from `spec.projectInfo`. That workspace has no program by construction, and the destroy runs from backend state, so installing can only fail or waste time.

The operator can decide this locally — `spec.projectInfo` is on the Workspace it is reconciling — with no RPC and no agent involvement.

## Note

This is not fixed incidentally by skipping install for prebuilt binaries (#1297): `writeProjectFile` passes `nil` runtime options, so the synthesized project drops `runtime.options.binary` even for a project that declared it. Detection based on the project file can never fire on this path.

## Affected product version(s)

v2.9.0 (the project-info destroy path was added in v2.8.0 via #1222).

guineveresaenger:
When we're running a project-info-bootstrapped workspace, the idea is to run `pulumi destroy` from project state only. If we try to run the Install step on the stack, this results in failures, since when there's no program in the Workspace, there's nothing to install.

Fixes https://github.com/pulumi/pulumi-kubernetes-operator/issues/1299.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 ## Unreleased
 
 - Skip `pulumi install` for a project that runs a prebuilt binary (`runtime.options.binary`) [#1297](https://github.com/pulumi/pulumi-kubernetes-operator/issues/1297)
+- Don't run `pulumi install` on a workspace bootstrapped from project info [#1299](https://github.com/pulumi/pulumi-kubernetes-operator/issues/1299)
 
 ## 2.9.0 (2026-07-29)
 
--- a/operator/internal/controller/auto/workspace_controller.go
+++ b/operator/internal/controller/auto/workspace_controller.go
@@ -314,28 +314,32 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			}
 		}
 
-		l.Info("Running pulumi install")
-		ready.Status = metav1.ConditionFalse
-		ready.Reason = "Installing"
-		ready.Message = "Installing packages and plugins required by the program"
-		if err := updateStatus(); err != nil {
-			return ctrl.Result{}, err
-		}
-		_, err = wc.Install(ctx, &agentpb.InstallRequest{})
-		if err != nil {
-			l.Error(err, "unable to install; marking workspace as stalled")
-			emitEvent(r.Recorder, w, autov1alpha1.InstallationFailureEvent(), err.Error())
+		if w.Spec.ProjectInfo != nil {
+			l.Info("Skipping pulumi install; the workspace has no program")
+		} else {
+			l.Info("Running pulumi install")
 			ready.Status = metav1.ConditionFalse
-			ready.Reason = "InstallationFailed"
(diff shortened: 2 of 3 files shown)


guineveresaenger:
@pulumi-eon please review

pose (approved):

PR #1327, opened 2026-09-03
Fix duplicative init container on no-source destroy path

guineveresaenger:
This pull request fixes a bug where, during no-source destroy, if the user's Workspace template had a Source set, the reconciler patched the Workspace to contain _two_ sources (`projectInfo` and what the user set).

When `projectInfo` is set as the Source on the Workspace, during template-patch apply, we unset all sources on the template, because the Workspace can only have one Source. The remainder of the template is merged as before.

Additionally, this contains a refactor where we add `projectInfo` to the full Source switch in the reconciler.

Tests for both changes were amended/added.


guineveresaenger:
@pulumi-eon please review

guineveresaenger:
@pulumi-eon can you fix that test nit?

pose (approved):

